PR #12, opened 2026-01-29
Fix crypto compatibility and timezone handling

AntTheLimey:
## Summary

- **Crypto compatibility**: Collector can decrypt passwords using both the server's random-salt format and the legacy username-salt format
- **Timezone fix**: The collector was storing timestamps in local time while the server compared against UTC. This caused two issues:
  - Servers showed as "Initializing" instead of "Online" (status calculation compared current UTC time against local-time timestamps, making recent data appear stale)
  - All time-based client views were offset by the user's timezone difference from UTC (e.g., a user in EST would see data appear 5 hours older than it actually was, affecting metrics, timeline events, and any time-range queries)
- **Gitignore**: Added `server/data/` to prevent committing runtime auth databases

## Test plan

- [ ] Verify servers show correct status (Online/Warning/Offline) regardless of user timezone
- [ ] Verify timeline events and metrics display with correct timestamps
- [ ] Verify collector can connect to databases with passwords encrypted by the server

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved password decryption to support both legacy and current encryption formats with automatic fallback.
  * Fixed timestamp calculations to use UTC consistently across metrics and server status tracking.
  * Enhanced configuration management to properly merge connection security settings.

* **Chores**
  * Updated ignore patterns for runtime data directories.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,9 @@ password.txt
 *.tmp
 tmp/
 temp/
+
+# Runtime data directories
+server/data/
 coverage.html
 alerter/ai-dba-alerter-linux
 CLAUDE.local.md
--- a/collector/src/database/crypto.go
+++ b/collector/src/database/crypto.go
@@ -74,43 +74,85 @@ func EncryptPassword(password string, serverSecret string, username string) (str
 	return base64.StdEncoding.EncodeToString(ciphertext), nil
 }
 
-// DecryptPassword decrypts a password using AES-256-GCM
-// The key is derived from the server secret and username
+// DecryptPassword decrypts a password using AES-256-GCM.
+// Supports two formats:
+// 1. New format (server): [salt (16 bytes)][nonce][ciphertext] - random salt
+// 2. Legacy format (collector): [nonce][ciphertext] - username as salt
 func DecryptPassword(encryptedPassword string, serverSecret string, username string) (string, error) {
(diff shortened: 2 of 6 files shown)


dpage:
Hi,

On the crypto side, this came about because the encryption mechanism changed, but not entirely consistently. Instead of adding legacy support to the collector we should:

  1. Update the collector's EncryptPassword to use the server format (random salt).                                                                                                             
  2. Remove all legacy format decryption paths from both collector and server.                                                                                                                  
  3. Extract a shared crypto package to eliminate the duplication of code between components.
  
Regarding the timestamp issue, the fix should be primarily made in the database. We need a migration that ensures every single timestamp in the database schema is a timestamptz (timestamp with timezone). That will give us absolute certainty on the precise times in the database, and from there we can ensure the other components handle that as expected.

Please close this PR, and re-submit as 2 separate ones, one for each concern. I'll handle the .gitignore update.

Thanks!

AntTheLimey:
Closing as requested. Will resubmit as two separate PRs with the recommended approaches:
1. Crypto: Standardize on server format with shared crypto package
2. Timestamps: Database migration to use timestamptz

PR #231, opened 2026-05-13
perf(collector): datastore schema and probe efficiency improvements (#219)

### LINKED ISSUE #219: Review index/schema design, based on findings from test deployment
# ai-workbench Database — Performance Review

**Database:** `ai_workbench` on `ai-workbench`
**Date:** 2026-05-12
**Scope:** Read-only inspection of `pg_stat_user_tables`, `pg_stat_user_indexes`, `pg_statio_user_tables`, `pg_stat_activity`, and partition metadata.

---

## Summary

162 user tables across `public` (application data) and `metrics` (weekly-partitioned Postgres telemetry). Total ~60+ GB. One application table (`anomaly_candidates`) is responsible for the bulk of the I/O pressure; the metrics ingestion pipeline is over-collecting `pg_settings` and carrying ~5 GB of unused indexes.

---

## 🔴 1. `anomaly_candidates` is being scanned to death

This is the standout problem.

- **669k sequential scans**, reading **112 billion tuples** cumulatively
- Table is 634 MB / ~2M rows
- 1.06 billion heap blocks read from disk vs 4.7 billion from cache — **~9.1 TB of physical I/O** on this one table
- Indexes exist but are barely used:

| Index | Scans |
|---|---|
| `anomaly_candidates_pkey` | 32,937,591 |
| `idx_anomaly_candidates_detected` | 11,622 |
| `idx_anomaly_candidates_connection` | 1 |
| `idx_anomaly_candidates_lookup` | 0 |

A query pattern is repeatedly scanning the whole table when it shouldn't be. Either the existing indexes don't match the predicate, or the planner thinks a seq scan is cheaper (often a stats / cost issue with bloated estimates). Worth `EXPLAIN`ing the actual query — find it in `pg_stat_statements`.

## 🟡 2. `alerts` — 2.2M seq scans on 51k rows

- 7.2 billion tuples sequentially read
- 1.63M UPDATEs, 1.56M HOT updates (95% HOT — that part's healthy)
- 6 indexes; primary key gets used heavily but seq scans still happen 2.2M times
- 2.5% dead tuples — autovacuum is keeping up

51k rows is small enough that the planner might choose seq scan over an index when it shouldn't. Or there's a query without a useful predicate (e.g., dashboard polling that filters in app code).

## 🟡 3. `metric_baselines` — mostly fine, occasional seq scans

7,033 rows, 12k full scans, but **39.8M index scans** — mostly using the index. There's a hot-path query that occasionally bypasses it. Probably benign.

## 🟡 4. ~5 GB of completely unused indexes on metrics partitions

Several large indexes have never been scanned:

| Index | Size | Scans |
|---|---|---|
| `pg_stat_all_indexes_20260504_connection_id_database_name_sc_idx` | 1687 MB | 0 |
| `pg_stat_all_indexes_20260504_pkey` | 1002 MB | 0 |
| `pg_stat_statements_20260504_pkey` | 549 MB | 0 |
| `pg_stat_statements_20260504_connection_id_database_name_que_idx` | 517 MB | 0 |
| `pg_stat_all_indexes_20260511_connection_id_database_name_sc_idx` | 398 MB | 0 |
| `pg_stat_statements_20260511_pkey` | 144 MB | 0 |
| `pg_stat_statements_20260511_connection_id_database_name_que_idx` | 114 MB | 0 |

The metrics data is being collected but never queried — these partitions are effectively write-only. Either drop the non-PK indexes from older partitions once they age out of query range, or stop creating them in the first place if nothing reads them.

## 🟡 5. `pg_settings` partition is the wrong shape

The `pg_settings_20260504` partition (one week) holds:

- **2,567 distinct collection timestamps**
- **~394 rows per collection** (441 settings × 17 connections, minus a few)
- **~1M rows total**
- **470 MB on disk**

That's roughly **150 collections per minute**. `pg_settings` is one of the least-changing things in Postgres — daily snapshots, or change-only capture, would save several GB per week plus the associated write I/O. There are 15 weekly `pg_settings` partitions in the metrics schema right now, each in the 100-500 MB range.

## 🟢 6. `anomaly_embeddings` index is huge but probably OK

30 GB total, 162 MB heap, ~30 GB index. That's a pgvector HNSW or IVFFlat index doing its job — those are inherently large. The only flag is that index maintenance during inserts is non-trivial; if ingest feels slow, the embedding index is the suspect.

## 🟢 7. `connections` seq scans — not a real problem

31M seq scans sounds alarming, but the table has 17 rows and fits in a single 8 KB page, so a seq scan there is literally one heap fetch. The planner is doing the right thing. Cache hit rate ~99.9%.

## 🟢 8. Partitioning is set up correctly

All the `metrics.pg_*` parents are declarative partitions (`relkind = 'p'`) with proper children — not legacy inheritance. No issue there.

## 🟢 9. No active long-running queries / lock waits

Snapshot of `pg_stat_activity` was clean at the time of inspection.

---

## Top three actions

1. **Find the query that's seq-scanning `anomaly_candidates`.** Query `metrics.pg_stat_statements_20260511` for statements touching that table, then EXPLAIN the worst one. The unused `idx_anomaly_candidates_lookup` and `_connection` indexes strongly hint that the query predicate doesn't match what was indexed for.
2. **Identify why `alerts` gets 2.2M seq scans.** Same approach via `pg_stat_statements`. Likely a poll without a selective `WHERE` clause, or a missing partial index on `(status, triggered_at)` or similar.
3. **Audit metrics collection cadence.** At minimum, drop `pg_settings` collection to daily or change-only, and prune unused non-PK indexes on older weekly partitions.

dpage:
## Summary

Three independent datastore performance fixes identified by the
production review in issue #219, bundled into one collector
migration v4 plus a probe fix.

- **Anomaly_candidates seq-scan**. Add a partial index on
  `anomaly_candidates(detected_at)` filtered to
  `processed_at IS NULL AND tier1_pass = TRUE`, so the alerter
  sweeper stops sequential-scanning the ~2M-row table on every
  poll. The partial WHERE matches the predicate exactly and the
  leading column matches the ORDER BY.
- **Redundant metrics indexes**. Drop the
  `idx_pg_stat_all_indexes_conn_db_time` and
  `idx_pg_stat_statements_conn_db_time` parent indexes; each is
  fully covered by the more specific `_object` index on the same
  table. The DROP cascades to every attached child-partition
  index automatically because partitioned-index children are
  owned by the parent. The v1 migration no longer creates these
  indexes on fresh installs.
- **`pg_settings` over-collection (~15x)**. Live probe metrics
  produced via `WrapQuery` + `ScanRowsToMaps` carry an extra
  `ai_dba_wb_probe` column that the stored-fetch query does not
  project. Without stripping it, the live and stored hashes
  never matched and `HasDataChanged` wrote a fresh snapshot
  every collection cycle. `HasDataChanged` now strips the marker
  before hashing, which also fixes the same latent bug on
  `pg_extension`, `pg_hba_file_rules`, and
  `pg_ident_file_mappings`.

## Test plan

- [x] `cd collector && make test-all` passes (89.0% on the
      probes package, 81.8% on the database package; touched
      functions `stripProbeMarker` and `normalizeDatabaseName`
      at 100%, `HasDataChanged` at 90.0%).
- [x] `cd alerter && make test-all` passes (unchanged code; the
      partial index simply unlocks the existing sweeper query).
- [x] `cd server && make test-all` passes (unchanged code).
- [x] `cd client && npm run test` passes (3104 tests; unrelated).
- [x] `TestMigrationV4_AnomalyCandidatesPartialIndexUsedByPlanner`
      verifies the planner picks the new partial index via
      `EXPLAIN`.
- [x] `TestMigrationV4_RedundantMetricsIndexesDropped` exercises
      the upgrade path: recreate the legacy partitioned indexes
      on a freshly migrated database with an existing weekly
      partition, then verify the DROP cascades to every attached
      child index.
- [x] `TestMigrationV4_Idempotent` confirms re-running the
      migration is a no-op.
- [x] `TestPgSettingsProbe_ExecuteAndStore` strengthened to assert
      `COUNT(DISTINCT collected_at) == 1` after two identical
      Store() calls, catching the regression we just fixed.
- [x] `TestStripProbeMarker_FixesOverCollectionHashMismatch` is a
      direct unit-test reproduction of the hash mismatch and the
      fix.

Fixes #219.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary snapshot churn by stripping the synthetic probe marker before metrics hashing.

* **Performance**
  * Added a targeted partial index to speed anomaly-candidate queries.
  * Removed redundant partitioned metrics indexes to reduce storage and maintenance.

* **Tests**
  * Added migration v4 tests (index presence, planner use, index removals, idempotency) and probe change-detection tests; improved integration test isolation.

* **Documentation**
  * Updated changelog with these datastore and probe improvements.

[Review Change Stack](https://app.coderabbit.ai/change-stack/pgEdge/ai-dba-workbench/pull/231)

## Codacy false positives

Codacy flags one (or two, depending on which static-analysis pass is
applied) "critical" security issue on this PR which its own summary
explicitly labels as a false positive:

> `2` new issues
> Security | `2` critical (2 false positives)

The flagged sites are dynamic identifiers in **test code** (e.g.
`fmt.Sprintf("CREATE INDEX %s ON %s %s", indexIdent, tableIdent, expr)`
inside `collector/src/database/migration_v4_test.go`) where every
input is either a local string constant or already routed through
`pgx.Identifier.Sanitize()`. The same pattern is used in production
code (`collector/src/probes/partition.go::dropExpiredPartitions`)
without issue. There is no user-controlled input on any of the
flagged paths.

These should be dismissed in the Codacy UI rather than worked around
in code. Restructuring to silence the rule would either obscure the
test intent or require duplicating identifiers in ways that defeat
the static guarantees `pgx.Identifier.Sanitize()` provides.

### DIFF
--- a/collector/src/database/migration_v4_test.go
+++ b/collector/src/database/migration_v4_test.go
@@ -1,0 +1,477 @@
+/*-------------------------------------------------------------------------
+ *
+ * pgEdge AI DBA Workbench
+ *
+ * Copyright (c) 2025 - 2026, pgEdge, Inc.
+ * This software is released under The PostgreSQL License
+ *
+ *-------------------------------------------------------------------------
+ */
+package database
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// TestMigrationV4_AnomalyCandidatesPartialIndexExists verifies that the v4
+// migration creates the partial index used by the alerter sweeper query.
+// The fixture migrates a fresh test database to the latest schema version
+// and asserts the index exists with the expected predicate and key column.
+func TestMigrationV4_AnomalyCandidatesPartialIndexExists(t *testing.T) {
+	ctx := context.Background()
+	pool, conn := getTestConnection(t)
+	if pool == nil {
+		return
+	}
+	defer pool.Close()
+	defer conn.Release()
+
+	cleanupTestSchema(t, pool)
+	defer cleanupTestSchema(t, pool)
+
+	sm := NewSchemaManager()
+	if err := sm.Migrate(conn); err != nil {
+		t.Fatalf("Failed to migrate: %v", err)
+	}
+
+	var indexDef string
+	err := pool.QueryRow(ctx, `
+		SELECT indexdef
+		FROM pg_indexes
+		WHERE schemaname = 'public'
+		  AND tablename = 'anomaly_candidates'
+		  AND indexname = 'idx_anomaly_candidates_unprocessed'
+	`).Scan(&indexDef)
+	if err != nil {
+		t.Fatalf("idx_anomaly_candidates_unprocessed missing: %v", err)
+	}
+
+	// The partial WHERE clause must match the alerter query predicate
+	// exactly so the planner can use the index. The leading sort column
+	// must be detected_at ASC so the ORDER BY can be served by the
+	// index without an extra sort node.
+	if !strings.Contains(indexDef, "detected_at") {
+		t.Errorf("index %s does not key on detected_at: %s",
+			"idx_anomaly_candidates_unprocessed", indexDef)
+	}
+	if !strings.Contains(indexDef, "processed_at IS NULL") {
+		t.Errorf("index %s missing processed_at IS NULL predicate: %s",
+			"idx_anomaly_candidates_unprocessed", indexDef)
+	}
+	if !strings.Contains(indexDef, "tier1_pass") {
+		t.Errorf("index %s missing tier1_pass predicate: %s",
+			"idx_anomaly_candidates_unprocessed", indexDef)
+	}
+}
+
+// TestMigrationV4_AnomalyCandidatesPartialIndexUsedByPlanner asserts that
+// the partial index is actually picked by the planner for the alerter's
+// hot GetUnprocessedAnomalyCandidates query. A seq-scan plan here would
+// silently re-introduce the regression we fixed.
+func TestMigrationV4_AnomalyCandidatesPartialIndexUsedByPlanner(t *testing.T) {
+	ctx := context.Background()
+	pool, conn := getTestConnection(t)
+	if pool == nil {
+		return
+	}
+	defer pool.Close()
+	defer conn.Release()
+
+	cleanupTestSchema(t, pool)
+	defer cleanupTestSchema(t, pool)
+
+	sm := NewSchemaManager()
+	if err := sm.Migrate(conn); err != nil {
+		t.Fatalf("Failed to migrate: %v", err)
+	}
+
+	// Insert a representative connection plus a small fixture of rows so
+	// the planner has stats to reason about. Anomaly_candidates depends
+	// on a connections row via FK; the connections row needs
+	// owner_username (the CHECK constraint requires exactly one of
+	// owner_username or owner_token).
+	_, err := pool.Exec(ctx, `
+		INSERT INTO connections (owner_username, name, host, port, database_name, username)
+		VALUES ('plan-fixture-user', 'plan-fixture', 'localhost', 5432, 'postgres', 'postgres')
+		ON CONFLICT DO NOTHING
+	`)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


dpage:
@coderabbitai review

PR #297, opened 2026-06-12
Migrate all LLM interface code to pgedge-go-llm-lib v0.1.0 (Phases B-E)

dpage:
## Summary

This migrates **all** LLM interface code in the Workbench onto the shared
library `github.com/pgEdge/pgedge-go-llm-lib` (pinned at `v0.1.0`), removing
the bespoke per-provider HTTP clients, the hand-written LLM proxy, and the
duplicated retry/embedding/logging code. Net result: ~6,600 fewer lines, one
tested implementation of LLM handling shared with the rest of pgEdge, and no
loss of Workbench functionality.

It covers four migration phases (B–E in the design spec); Phase A — the
upstream library work and the `v0.1.0` release — landed separately in the
library repo.

## What changed, by phase

**B — Embeddings (`pkg/embedding`):** the five bespoke HTTP embedding
providers are replaced by a thin wrapper over `llm.Client`. The `Provider`
interface, `Config`, and constructor signatures are preserved; the
construction-time model allow-list (OpenAI/Voyage/Gemini) is retained for
no-loss behaviour and KB vector compatibility. Consumers use the upstream
`llm/vec` conversion helpers.

**C — Alerter (`alerter/internal/llm`):** the four bespoke reasoning
providers and the hand-rolled `retry.go` are replaced by one `libReasoning`
wrapper over `llm.Client.Chat`; the embedding adapter uses `llm/vec` for
resize/normalise. The `EmbeddingProvider`/`ReasoningProvider` interfaces and
the engine are unchanged.

**D — Server chat + proxy:** the server adopts the library's `llm/proxy`
gateway behind a thin shim mounted at `/api/v1/llm`. A shared
`auth.AuthenticateRequest` backs both the HTTP middleware and the proxy
`Authorize` hook (public `providers`/`models`/`health`; authed
`chat`/`chat/stream`/`embed`/`rerank`). A `TransformRequest` hook carries the
existing server-side injection: the default system prompt, pinned-memory and
user/RBAC context (same order and caps), compact tool descriptions, and
provider headers. The overview generator and server-info AI analysis move to
`llm.Client` directly. The bespoke chat clients, factory, wire types,
transport, and the now-callerless `pkg/embedding/logger.go` are deleted.

**E — React client:** updated to the library proxy wire contract
(`system`→`system_prompt`, `token_usage`→`usage`, message content always
typed blocks, nested `tool_use`, `tool_result.text`). A single
`normaliseMessages` boundary guarantees no bare-string content reaches the
wire; the agentic tool-calling loop and all single-shot analyses are
preserved. The chat URL is unchanged.

## Breaking change (server LLM wire contract)

`POST /api/v1/llm/chat` now uses the library shape: request `system_prompt`
(was `system`) and typed-block message content; response `usage` (was
`token_usage`) and nested `tool_use` content blocks. The React client in this
PR is updated in lockstep, and the OpenAPI spec + reference table are
regenerated. New endpoints `/chat/stream`, `/embed`, `/rerank`, `/health` are
exposed (authed except `/health`).

## Security

The proxy shim's auth and injection paths were reviewed by a security audit.
Auth was **not** weakened (parity with the prior implementation: same
token-then-session validation, context keys, and 401s; no public-path bypass;
per-user memory scoping preserved). One finding was fixed: model-name
validation was restored (and hardened to reject `..`) to block
authenticated upstream request-path manipulation against the Gemini provider.

## Testing

`make test-all` is green: all Go modules (build, vet, race tests, lint, ≥90%
coverage on touched units) against the local Postgres, plus the full client
vitest suite (3353 tests) and `tsc`/vite build. Every phase was implemented
test-first with spec-compliance and code-quality review.

## Follow-ups (non-blocking)

- A manual `playwright-cli` chat session against a running dev server built
  from this branch (the dev server runs from the main checkout, so this is a
  post-merge check).
- Optional: bound the inherited retry policy for the background analysis
  consumers; consider disabling `/embed/multimodal` if unused.

Depends on `github.com/pgEdge/pgedge-go-llm-lib v0.1.0`.

## Summary by CodeRabbit

* **New Features**
  * Added a unified LLM proxy route with model-name safety checks and system-prompt/tool injection
  * Introduced `/llm/chat/stream` (SSE), `/llm/embed` (text + multimodal), and `/llm/rerank`, plus public `/llm/providers`, `/llm/models`, and `/llm/health`
  * Standardized chat messaging with typed content blocks and updated tool/result payloads
* **Bug Fixes**
  * Centralized authentication for LLM routes and improved auth/error handling
* **Chores**
  * Updated Go toolchain versions to 1.26.1
  * Consolidated LLM/embedding logic into the shared library layer
* **Tests**
  * Refreshed and expanded unit/integration coverage for the new proxy, payload contract, and embedding/tool flows


dpage:
Thanks @coderabbitai — addressed in `4509e73` and `9d6c9a2`:

- **5 MB body-cap bypass (proxy.go)** ✅ fixed: the model-validation middleware now reads `maxChatBodySize+1` and rejects an over-cap body with 413 instead of silently truncating and processing the prefix. Tests added for over-cap (413) and at-cap (passes through intact).
- **`TimeoutSeconds` ignored in the direct builders** ✅ fixed: both `overview.createLLMClient` and `server_info.createLLMClient` now set `RequestTimeout` from `LLMConfig.TimeoutSeconds` (guarded `> 0`), matching the proxy shim. Tests added.
- **nil `g.llmConfig` deref (overview)** ✅ fixed: `createLLMClient` returns a "no LLM provider configured" error before any dereference; the summary/refresh paths already handle that. Test added. (server-info already guards before calling.)
- **Embedding DB test isolation** ✅ fixed: the test now creates a uniquely-named private schema with its own `chat_memories vector(3)` and drops it in teardown; the shared `public.chat_memories` is never touched.
- **SSE `error` event in the stream schema** ✅ fixed, with a correction: the library emits the error as a *separate* SSE event (`event: error\ndata: {"error":"..."}`), not a stream chunk with `type:"error"`. So rather than adding `error` to the chunk `type` enum (which would misdescribe the wire), I documented the terminal error event and an optional `error` field on the stream schema. OpenAPI regenerated; spec tests pass.
- **`/embed/multimodal`** ✅ added to both the OpenAPI spec and the `reference.md` summary table.
- **Copyright header** ✅ added to `pkg/embedding/libprovider_test.go` (the other new test files already had it; the repo convention does apply to test files).

- **`LLMContentBlock` "missing nested `tool_result`"**: respectfully, this one I don't think is a defect. In the library wire format only `tool_use` is nested (`{type:"tool_use", tool_use:{...}}`); `tool_result` is **flat** — `{type:"tool_result", tool_use_id, text, is_error}` — both in `llm.ContentBlock` and in what the migrated client sends. So the schema documenting `tool_result` with flat `text`/`tool_use_id`/`is_error` matches the actual contract. Leaving it as-is.

`make test-all` is green (all Go modules + the full client suite).

dpage:
Follow-up on the `memory_tools_embedding_db_test.go` comments (fixed in `7666f2d`):

- **Schema isolation:** the test no longer touches the shared `public.chat_memories`. It creates a uniquely-named private schema (`pgx.Identifier`-quoted), builds its own `chat_memories vector(3)` inside it via `search_path`, and drops the whole schema in teardown. No shared relation is dropped or altered.
- **`CREATE EXTENSION ... SCHEMA public`:** applied, so the `vector` extension persists in `public` rather than being created in (and dropped from) the private schema each run.
- **Codacy SQL-injection findings:** these were false positives (test-only DDL; the schema name is a sanitised, non-user-controlled test name). Resolved by quoting the dynamic identifiers with `pgx.Identifier{...}.Sanitize()` (the idiomatic safe form), leaving a single justified `//nosemgrep` on the constant `CREATE EXTENSION` literal. Verified locally with the Codacy CLI: Opengrep, Lizard, and Trivy all report zero issues on the file.

Also reduced `transformRequest` and `isValidModelName` cyclomatic complexity below the threshold (extracted injection helpers; regexp-based validator).

server/src/internal/llmproxy/proxy.go
```diff
@@ -44,389 +62,377 @@ type Config struct {
...
-		}
-		http.Error(w, "LLM request failed", http.StatusInternalServerError)
+// onResponse logs the assembled LLM response when tracing is enabled,
+// reproducing the old HandleChat LogLLMResponse call.
+func (c *Config) onResponse(r *http.Request, info proxy.ResponseInfo) {
+	if !tracing.IsEnabled() || info.Response == nil {
 		return
 	}
+	tokenHash := auth.GetTokenHashFromContext(r.Context())
+	tracing.LogLLMResponse(tokenHash, tokenHash, info.RequestID, info.Response.Content, 0)
```
tsivaprasad: It looks like the last parameter (duration) is always being passed as 0. Could we pass the actual duration instead?

server/src/internal/api/server_info_handlers.go
```diff
@@ -815,70 +809,54 @@ func parseDatabaseAnalysisResponse(
...
 
-// createLLMClient builds the appropriate chat client based on the
-// configured LLM provider.
-func (h *ServerInfoHandler) createLLMClient() chat.LLMClient {
-	switch h.llmConfig.Provider {
+// createLLMClient builds a pgedge-go-llm-lib client based on the
+// configured LLM provider. It selects the per-provider API key, model,
+// and base URL from the injected LLM configuration and returns the
+// construction error so callers can skip analysis on misconfiguration.
+func (h *ServerInfoHandler) createLLMClient() (pgllm.Client, error) {
```
tsivaprasad, line 816: We can simplify and refactor this function code and [here](https://github.com/pgEdge/ai-dba-workbench/pull/297/changes#diff-ea0ba55d98e4468036af7b95f7f1d1c1851a6cfaf1d39782fe3544fc17a93143R530)

tsivaprasad (changes requested):

server/src/internal/api/server_info_handlers.go
dpage: Agreed, thanks. Done in 40426af4: the shared logic now lives in a single exported helper `(*llmproxy.Config).BuildClientOptions(maxTokens, temperature)` in `llmproxy`, which owns the per-provider credential selection, the `pgllm.Options` construction, the timeout-only-when-positive rule, and the custom-header wiring. Both `ServerInfoHandler.createLLMClient` and `Generator.createLLMClient` now call it and keep only their own per-site `NewClient` error logging and (for the generator) the nil/empty-provider guard; the now-unused per-site `getProviderHeaders` methods are removed as dead code. It is a behaviour-preserving refactor, and the touched functions are at 100% line coverage.

docs/admin-guide/api/openapi.json
```diff
@@ -4796,10 +4796,168 @@
...
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Server-Sent Event stream of chat chunks",
+            "content": {
+              "text/event-stream": {
+                "schema": {
+                  "$ref": "#/components/schemas/LLMStreamChunk"
+                }
```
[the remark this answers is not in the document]
dpage: Having checked this against the library's actual SSE implementation (`llm/proxy/sse.go` and `types.go`), the current schema is already correct and the suggested change would misdescribe the wire. A streamed error is not a `StreamChunk` with `type: "error"`; it is emitted as a distinct terminal SSE frame, `event: error\ndata: {"error":"..."}`, whose payload is an `ErrorResponse` object, not a `StreamChunk`. The `llm.StreamChunk` struct has no error member and no `"error"` enum value (its fields are `type`, `text`, `tool_use`, `partial`, `usage`). The `LLMStreamChunk` schema therefore correctly keeps the enum as `text|tool_use_start|tool_use_delta|done`, and the terminal `event: error` payload is already documented in prose plus an `error` string property. `make openapi` produces a zero diff, so source and spec are in sync. Resolving as already addressed.

docs/admin-guide/api/openapi.json
```diff
@@ -10687,6 +10987,175 @@
...
+              "document",
+              "tool_use",
+              "tool_result"
+            ]
+          }
+        },
+        "required": [
+          "type"
+        ]
+      },
```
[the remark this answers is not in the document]
dpage: Checked against the library's `llm.ContentBlock` struct and the React client, and the schema already matches the real wire contract; the nested `tool_result` shape suggested here does not exist on the wire. A `tool_result` block is flattened, not nested: `llm.ContentBlock` serialises `type`, `text` (used for both text and simple tool_result), `image`, `document`, `tool_use` (the only nested object), `tool_use_id`, `is_error`, and `cache_control` — there is no `tool_result` object and no `tool_result.text`. The React side agrees: `client/src/types/llm.ts` defines `ToolResult` as `{type, tool_use_id, text, is_error?}` (flattened), and `normaliseMessages` passes it through unchanged. The current `LLMContentBlock` schema already has `type` required, the enum `text|image|document|tool_use|tool_result`, the nested `tool_use`, and the flattened `text`/`tool_use_id`/`is_error` fields, so it is accurate as-is. Resolving as already addressed.

server/src/internal/llmproxy/proxy.go
dpage: Good catch, this is a genuine loss of parity: the pre-migration `HandleChat` timed the call and logged the real elapsed duration, whereas the current hook hardcodes `0`. Rather than reconstruct the timing locally in the proxy shim (which would mean correlating `onRequest`/`onResponse` via a `sync.Map` and can't see the true upstream boundary), we're fixing it at the right layer: the library proxy owns the upstream call and can expose the exact duration to every consumer. That change is up as pgEdge/pgedge-go-llm-lib#20, which adds a `Duration` field to the proxy's `ResponseInfo`/`ErrorInfo` (backward compatible with v0.1.0, additive struct fields only).

Once that's released as v0.1.1, this PR will bump `go.mod` and the `onResponse` hook here will pass `info.Duration` through to `LogLLMResponse`. Leaving this thread open until that bump lands.

dpage: Done in bc704965. pgEdge/pgedge-go-llm-lib was released as v0.1.1 with the `Duration` field on the proxy's `ResponseInfo`/`ErrorInfo`, so this PR now pins v0.1.1 (bumped across the server, alerter, and pkg modules) and the `onResponse` hook forwards `info.Duration` into `LogLLMResponse` in place of the old hardcoded `0`. The library proxy measures the upstream provider call and hands over its wall-clock duration, restoring the response-timing telemetry the pre-migration `HandleChat` recorded. `onError` is left as-is because `LogError` takes no duration parameter. A test asserts the forwarded value now appears as `"duration_ms":1500` rather than zero. Resolving.

server/src/internal/tools/memory_tools_embedding_db_test.go
```diff
@@ -0,0 +1,262 @@
...
+		t.Skipf("pgvector extension unavailable: %v", err)
+	}
+	// Quote the schema name via pgx.Identifier so it is safely escaped; the
+	// name itself is derived from the SQL-safe test name, not user input.
+	schemaIdent := pgx.Identifier{schema}.Sanitize()
+	//nosemgrep: go_sql_rule-concat-sqli -- test-only DDL; schema identifier sanitized via pgx.Identifier, DDL identifiers cannot be parameterized
+	if _, err := pool.Exec(ctx,
+		fmt.Sprintf(`DROP SCHEMA IF EXISTS %s CASCADE; CREATE SCHEMA %s`, schemaIdent, schemaIdent)); err != nil {
+		pool.Close()
+		t.Skipf("Failed to create private test schema: %v", err)
```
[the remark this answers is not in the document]
dpage: I verified this against the running database and it's a false positive: the premise that schema setup fails and both integration tests skip does not hold.

pgx v5's `Conn.Exec` uses the simple query protocol when it is called with no query arguments, and the simple protocol fully supports multiple semicolon-separated statements. All three DDL `Exec` calls here (the `DROP SCHEMA ...; CREATE SCHEMA ...` on lines 106-107 included) pass no arguments, so they go through the simple protocol, not the extended/prepared-statement path that rejects multiple commands. The "cannot insert multiple commands into a prepared statement" error only arises on the extended protocol, which pgx uses for parameterised queries.

Empirically, against local PostgreSQL 18 + pgvector with pgx v5.9.2, both integration tests run and pass rather than skip:

```
=== RUN   TestStoreMemoryGeneratesEmbeddingIntegration
--- PASS: TestStoreMemoryGeneratesEmbeddingIntegration (0.02s)
=== RUN   TestRecallMemoriesGeneratesQueryEmbeddingIntegration
--- PASS: TestRecallMemoriesGeneratesQueryEmbeddingIntegration (0.02s)
PASS
ok  github.com/pgedge/ai-workbench/server/internal/tools  0.314s
```

A skipped setup would have surfaced as `--- SKIP` via the `t.Skipf` guard; both show `--- PASS` with real per-test DB work. The nested ast-grep SQL-injection note is likewise a non-issue: the schema identifier is sanitised via `pgx.Identifier{...}.Sanitize()` and derived from the SQL-safe test name, not user input, and DDL identifiers cannot be parameterised (already annotated accordingly). No code change needed; resolving as verified.

dpage:
@tsivaprasad this is ready for another look. All of your review feedback has been addressed and every thread is resolved:

- **`createLLMClient` duplication:** refactored into a shared `llmproxy.Config.BuildClientOptions` helper, both call sites updated, dead code removed.
- **`duration` always `0`:** fixed at the right layer. The upstream fix landed in `pgedge-go-llm-lib` v0.1.1 (exposes the upstream-call `Duration` on the proxy hook info), and this PR now pins v0.1.1 and forwards `info.Duration` into the tracer in place of the hardcoded `0`.

The two CodeRabbit items (streamed-error schema and nested `tool_result`) were verified against the real library/client wire types and resolved as already-correct. CI is green across the full matrix. Re-requesting your review; over to you when you have a moment.

tsivaprasad (approved):
Looks good.

PR #310, opened 2026-06-29
Update binary_install.md

paul-pgedge:
Clarified which login details to use when accessing the workbench for the first time based on user feedback

## Summary by CodeRabbit

* **Documentation**
  * Clarified the Workbench startup login steps.
  * Users are now explicitly instructed to sign in at `http://<server-ip>` using the username and password created via the server’s `-add-user` command.
  * The guidance is now more direct and easier to follow when first accessing the Workbench.

### DIFF
--- a/docs/getting-started/binary_install.md
+++ b/docs/getting-started/binary_install.md
@@ -528,8 +528,7 @@ sudo nginx -t
 sudo systemctl restart nginx
 ```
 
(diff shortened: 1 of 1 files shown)


dpage (approved):
Reviewed; clarification is correct and useful. I pushed a small follow-up fixing the login-step grammar, restoring 79-char wrapping, and correcting the flag to `-add-user` to match the server CLI.

PR #316, opened 2026-06-30
Remove runtime pgedge-ai-kb dependency

maqeel75:
Remove pgedge-ai-kb dependency because now user will install dependency based on model he/she selects.

## Summary by CodeRabbit

* **Chores**
  * Updated runtime dependencies for Debian and RPM builds of the server package.
  * Dropped the previous explicit dependency on the knowledgebase package while retaining the OpenSSL requirement.
* **Documentation**
  * Updated the changelog to reflect that knowledgebase is now delivered as opt-in, co-installable packages (configured by operators).

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,6 +8,19 @@ The format is based on
 project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Drop the runtime dependency on a knowledgebase package from the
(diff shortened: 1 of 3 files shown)


dpage (approved):
Reviewed end-to-end: removing the runtime `pgedge-ai-kb` dependency is correct, since the knowledgebase is now published as co-installable `pgedge-ai-kb-<provider>-<model>` packages (no single `pgedge-ai-kb` package to depend on) and is opt-in/disabled by default. Added an Unreleased changelog entry. CI, Codacy and CodeRabbit all green.

PR #317, opened 2026-06-30
fix(client): show group-inherited admin privileges on the Users tab

### LINKED ISSUE #296: Privileges not accurately displayed; ADMIN privileges omitted
Image
Image

In the attached screenshots, you can see that Alice has several ADMIN privileges (inherited through group membership), but they are not displayed under her name on the Users tab.

dpage:
## Summary

- On the Users tab, a user's ADMIN privileges **inherited through group membership** were not displayed under their name (per #296, Alice had several inherited ADMIN privileges that were omitted).
- Root cause is purely client-side: in `EffectivePermissionsPanel.tsx` the Admin permissions card was gated on `isSuperuser`, so a non-superuser whose admin permissions come from a group (where `isSuperuser` is `false` but `adminPermissions` is non-empty) never saw the Admin section. The server already returns the correct effective `admin_permissions` (including group-inherited ones) from `/api/v1/rbac/users/{id}/privileges`.
- The fix gates the Admin card on `isSuperuser || adminPermissions.length > 0`. The superuser path is byte-for-byte unchanged (superusers still receive the `"*"` wildcard and render as before), and a non-superuser with no admin permissions still shows no Admin card, matching the sibling cards' intent of not adding noise for the common case.

This is a **display-only** change; server-side enforcement and effective-privilege computation are untouched.

## Test plan

- [x] Replaced the test that codified the old behaviour with two tests: a non-superuser with inherited admin permissions now shows the Admin section and its chips; a non-superuser with no admin permissions shows no Admin card. Existing superuser/wildcard tests still pass (19/19).
- [x] `npm run lint` clean (no new warnings on the touched files).
- [x] `EffectivePermissionsPanel.tsx` at 100% line coverage (above the 90% floor).

Closes #296

### DIFF
--- a/client/src/components/AdminPanel/EffectivePermissionsPanel.tsx
+++ b/client/src/components/AdminPanel/EffectivePermissionsPanel.tsx
@@ -146,7 +146,7 @@ const EffectivePermissionsPanel: React.FC<EffectivePermissionsPanelProps> = ({
                     )}
                 </CategoryCard>
 
-                {isSuperuser && (
+                {(isSuperuser || (adminPermissions && adminPermissions.length > 0)) && (
                     <CategoryCard icon={AdminIcon} label="Admin" testId="admin-permissions-section">
                         {adminPermissions && adminPermissions.length > 0 ? (
                             <Box sx={{ display: 'flex', flexWrap: 'wrap', gap: 0.5 }}>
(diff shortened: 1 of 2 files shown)


tsivaprasad (approved):
Looks good

PR #326, opened 2026-07-02
chore(collector): harden shutdown diagnostics, timeouts and compose (#308)

### LINKED ISSUE #308: suspected collector/runtime issue after upgrade
After upgrading pgEdge AI DBA Workbench from `1.0.0-beta3` to `1.0.0`, the Dashboard stopped showing Top Queries and displayed the following message:

`No query statistics available. Is the pg_stat_statements extension installed?`

The target PostgreSQL server was checked and `pg_stat_statements` was working correctly. The monitoring user was also able to query `pg_stat_statements` successfully.

Evidence from the target PostgreSQL server:

```sql
SELECT count(*) FROM pg_stat_statements;
```

returned rows successfully, and Top Queries were visible manually using the monitoring user.

The actual issue was on the Workbench collector side. The `ai-dba-collector` container was entering a clean-exit restart loop. It was not crashing with an error, and it was not being killed by OOM. The container exited with `ExitCode=0`, and Docker restarted it because the restart policy was set to `unless-stopped`.

Observed evidence:

```text
RestartCount increased continuously
ExitCode=0
OOMKilled=false
```

Collector logs repeatedly showed:

```text
pgEdge AI DBA Workbench Collector v1.0.0 starting...
Datastore connection established
Probe scheduler started
Collector is running. Press Ctrl+C to stop.
```

Later, the collector also logged:

```text
Probe pg_stat_statements execution timed out for connection DWH_PRE (timeout: 60 seconds)
```

This caused the Dashboard to stop displaying Top Queries, even though `pg_stat_statements` itself was installed and working correctly on the monitored PostgreSQL server.

Resolution applied:

1. Enabled interactive mode for the collector container in `docker-compose.production.yml`:

```yaml
collector:
  stdin_open: true
  tty: true
```

2. Increased collector memory limit:

```yaml
deploy:
  resources:
    limits:
      cpus: '1'
      memory: 1G
```

3. Tuned the collector connection pool in `docker/config/ai-dba-collector.yaml`:

```yaml
pool:
  datastore_max_connections: 25
  datastore_max_idle_seconds: 300
  datastore_max_wait_seconds: 120
  max_connections_per_server: 1
  monitored_max_idle_seconds: 300
  monitored_max_wait_seconds: 300
```

4. Recreated the collector container:

```bash
docker compose -f examples/docker-compose.production.yml up -d --force-recreate collector
```

After applying these changes, the collector became stable:

```text
RestartCount=0
Status=running
OOMKilled=false
Tty=true
OpenStdin=true
```

The Dashboard started showing Top Queries again.

Conclusion:

This was not a PostgreSQL permission issue and not a missing `pg_stat_statements` extension issue. The root cause was the `ai-dba-collector` container entering a clean-exit restart loop after the upgrade to `1.0.0`, combined with `pg_stat_statements` probe timeout behavior. The issue was resolved by enabling `stdin_open`/`tty`, increasing collector memory, reducing monitored server concurrency, and increasing monitored query timeout settings.

dpage:
## Summary

Addresses #308, where the `ai-dba-collector` container entered a clean-exit (`ExitCode=0`) restart loop after the upgrade to 1.0.0, which stopped the Dashboard from showing Top Queries.

**Important framing:** investigation found the collector's signal handling is correct and nothing in its running path can force an exit (probe timeouts, pool exhaustion, and GC errors are all handled gracefully; the only `Fatal` calls are at startup and exit with code 1). `ExitCode=0` therefore means a signal was actually delivered to the process, and the *source* of that signal could not be determined from the codebase (it needs deployment data we don't have: Docker/runc versions, launch mode, whether the shutdown log line appeared). Rather than guess at a root cause, this PR ships low-risk hardening agreed with the maintainer and leaves the signal source as a documented known-unknown.

## Changes

- **Diagnosable shutdown**: `waitForShutdown` now returns the received signal and `main` logs *which* signal triggered shutdown, instead of exiting silently. If the loop recurs, the logs will show whether it was SIGTERM/SIGINT rather than being opaque.
- **Timeout headroom**: default `monitored_max_wait_seconds` raised 60 → 120. This value also bounds each probe's execution deadline, and the report observed `pg_stat_statements` timing out at 60s on a large monitored database.
- **Compose hardening** (`examples/docker-compose.production.yml` and `docker-compose.yml`): the collector service now runs an init as PID 1 (`init: true`) for robust signal handling, has a `stop_grace_period: 30s` for its multi-step graceful shutdown, and a modestly higher memory limit (256M → 512M, precautionary headroom; the report was **not** OOM). `stdin_open`/`tty` were **deliberately not** added — they mask the symptom and are wrong for a background daemon.
- **Config documentation**: the collector pool/timeout knobs are now spelled out with comments in the sample `docker/config/ai-dba-collector.yaml`, so operators can discover and tune them.

## Test plan

- [x] `waitForShutdown` returns the signal (`main_test.go` asserts SIGTERM); `config_test.go` asserts the new 120s default. `waitForShutdown` and `NewConfig` at 100% coverage.
- [x] `gofmt` clean, `go build ./...` and the collector unit tests pass (against local Postgres).
- [x] Both compose files and the config YAML parse cleanly (validated with a YAML parser; `docker` was unavailable on the dev host to run `docker compose config -q`).

## Known-unknown

The precise reason the collector process received a shutdown signal in the reporter's environment is not identified; this PR makes any recurrence diagnosable rather than silent, and does not claim to fix that underlying cause.

Closes #308

## Summary by CodeRabbit

* **New Features**
  * None

* **Bug Fixes**
  * Improved collector shutdown by logging the exact OS signal that triggered termination.
  * Increased the default maximum wait time for monitored connections (from 60s to 120s) to help avoid interruptions during clean-exit restart loops.
  * Hardened container shutdown behavior with a 30s stop grace period and Docker init enabled for better signal handling.

* **Documentation**
  * Updated the changelog and sample configuration to include the new connection timing options and revised deployment guidance.

### DIFF
--- a/collector/src/config.go
+++ b/collector/src/config.go
@@ -70,7 +70,7 @@ func NewConfig() *Config {
 			DatastoreMaxWaitSeconds: 60,
 			MaxConnectionsPerServer: 3,
 			MonitoredMaxIdleSeconds: 300,
-			MonitoredMaxWaitSeconds: 60,
+			MonitoredMaxWaitSeconds: 120,
 		},
 	}
 }
--- a/collector/src/config_test.go
+++ b/collector/src/config_test.go
@@ -37,8 +37,8 @@ func TestNewConfig(t *testing.T) {
 		t.Errorf("Expected default Pool.DatastoreMaxWaitSeconds to be 60, got %d", config.Pool.DatastoreMaxWaitSeconds)
 	}
 
-	if config.Pool.MonitoredMaxWaitSeconds != 60 {
-		t.Errorf("Expected default Pool.MonitoredMaxWaitSeconds to be 60, got %d", config.Pool.MonitoredMaxWaitSeconds)
+	if config.Pool.MonitoredMaxWaitSeconds != 120 {
+		t.Errorf("Expected default Pool.MonitoredMaxWaitSeconds to be 120, got %d", config.Pool.MonitoredMaxWaitSeconds)
 	}
 
 	if config.Pool.MaxConnectionsPerServer != 3 {
--- a/collector/src/main.go
+++ b/collector/src/main.go
@@ -112,9 +112,9 @@ func main() {
 	logger.Startup("Collector is running. Press Ctrl+C to stop.")
 
 	// Wait for shutdown signal
-	waitForShutdown()
+	sig := waitForShutdown()
 
-	logger.Startup("Shutdown signal received, stopping...")
+	logger.Startupf("Received signal %q, shutting down...", sig)
 
 	// Shutdown in proper order to ensure clean connection closure
 	// 1. Stop probe scheduler (no new probe queries)
@@ -230,9 +230,13 @@ func loadConfiguration() (*Config, error) {
 	return config, nil
 }
 
-// waitForShutdown waits for an interrupt signal
-func waitForShutdown() {
+// waitForShutdown waits for an interrupt signal and returns the
+// signal that was received, so the caller can log which signal
+// triggered the shutdown. A clean exit that was previously silent
+// now records the delivered signal, making a future recurrence of
+// the reported clean-exit restart loop diagnosable.
+func waitForShutdown() os.Signal {
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
-	<-sigChan
+	return <-sigChan
 }
--- a/collector/src/main_test.go
+++ b/collector/src/main_test.go
@@ -396,11 +396,10 @@ func TestMaybePrintSchemaVersion_WriteError(t *testing.T) {
 }
 
 func TestWaitForShutdown(t *testing.T) {
-	done := make(chan struct{})
+	got := make(chan os.Signal, 1)
 
 	go func() {
-		waitForShutdown()
-		close(done)
+		got <- waitForShutdown()
 	}()
 
 	// Give the goroutine time to install its signal handler before we
@@ -416,8 +415,10 @@ func TestWaitForShutdown(t *testing.T) {
 	}
 
 	select {
-	case <-done:
-		// ok
+	case sig := <-got:
+		if sig != syscall.SIGTERM {
+			t.Errorf("waitForShutdown returned %v, want SIGTERM", sig)
+		}
 	case <-time.After(2 * time.Second):
 		t.Fatal("waitForShutdown did not return after SIGTERM")
 	}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -41,7 +41,17 @@ services:
       - ./docker/config/ai-dba-collector.yaml:/etc/pgedge/ai-dba-collector.yaml:ro
       - ./docker/secret/ai-dba.secret:/etc/pgedge/ai-dba.secret:ro
       - ./docker/secret/pg-password:/etc/pgedge/pg-password:ro
+    # Run an init process as PID 1 for robust signal forwarding and
+    # zombie reaping.
+    init: true
+    # Give the collector time to close its connection pools during its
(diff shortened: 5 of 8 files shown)


dpage:
Re: CodeRabbit's "Linked Issues" warning that this PR doesn't add `stdin_open`/`tty`, a 1G memory limit, or the exact pool values from #308 — that divergence is intentional and was agreed with the maintainer.

Those items were the reporter's environment-specific workarounds, not a diagnosed fix:

- **`stdin_open`/`tty`**: deliberately omitted. They keep a TTY/stdin attached to mask a clean-exit symptom, which is inappropriate for a background daemon. Investigation showed the collector's signal handling is correct and nothing in its running path forces an exit, so `ExitCode=0` means a signal was delivered; the source is undetermined and documented here as a known-unknown. This PR instead makes any recurrence *diagnosable* (it now logs which signal triggered shutdown) rather than papering over it.
- **Memory 1G → 512M**: the report explicitly noted `OOMKilled=false`, so memory was not the cause. 512M is a modest precautionary headroom bump; 1G was not justified by evidence.
- **Specific pool values (e.g. `max_connections_per_server: 1`, `monitored_max_wait_seconds: 300`)**: those were the reporter's tuning for their large DWH, not sensible global defaults. We raised the shipped default `monitored_max_wait_seconds` from 60 to 120 (the value that also bounds probe execution, which is what timed out) and now document all pool/timeout knobs so operators can tune for their own environment.

So the PR intentionally implements the low-risk, generally-applicable subset and avoids the deployment-specific masks. No code change is planned in response to this warning.

tsivaprasad (approved):
Looks good and verified.

```
 src git:(fix/issue-308-collector-shutdown-hardening) go test . -run 'TestWaitForShutdown|TestNewConfig' -v

=== RUN   TestNewConfig
--- PASS: TestNewConfig (0.00s)
=== RUN   TestWaitForShutdown
--- PASS: TestWaitForShutdown (0.05s)
PASS
ok  	github.com/pgedge/ai-workbench/collector/src
```

Verify SIGTERM is now named in the shutdown log:
```
ai-dba-workbench git:(fix/issue-308-collector-shutdown-hardening) timeout 5 ./bin/ai-dba-collector --config=bin/ai-dba-collector.yaml > /tmp/collector_sigterm_test.log 2>&1 &
COLLECTOR_PID=$!
sleep 1
kill -TERM $COLLECTOR_PID
sleep 1
cat /tmp/collector_sigterm_test.log
[1] 55183
[1]  + 55183 done       timeout 5 ./bin/ai-dba-collector --config=bin/ai-dba-collector.yaml >  2>&1
2026/07/23 16:04:29 pgEdge AI DBA Workbench Collector v1.0.0 starting...
2026/07/23 16:04:29 Configuration loaded from: bin/ai-dba-collector.yaml
2026/07/23 16:04:29 Schema is up to date
2026/07/23 16:04:29 Datastore connection established
2026/07/23 16:04:29 Probe scheduler started
2026/07/23 16:04:29 Collector is running. Press Ctrl+C to stop.
2026/07/23 16:04:29 Received signal "terminated", shutting down...
2026/07/23 16:04:29 Stopping probe scheduler...
2026/07/23 16:04:29 Probe scheduler stopped
2026/07/23 16:04:29 Garbage collector stopped
2026/07/23 16:04:29 Collector stopped
```
Confirm the Docker Compose hardening actually takes effect
```
 ai-dba-workbench git:(fix/issue-308-collector-shutdown-hardening) POSTGRES_PASSWORD=postgres docker compose config > /tmp/compose_resolved.yml
sed -n '/^  collector:/,/^  [a-z]/p' /tmp/compose_resolved.yml | grep -E "memory:|init:|stop_grace_period:"

          memory: "536870912"
    init: true
    stop_grace_period: 30s
```

PR #327, opened 2026-07-02
fix(client): resolve connection names in admin privilege chips

### LINKED ISSUE #309: Connection ID exposed in Group Privileges panel
**Component:** Admin → Groups → expanded group row → connection privilege chips

**Observed behavior:** Connection privileges display as Connection 3 (read) using the internal numeric database ID.

**Expected behavior:** Should display the human-readable connection name, e.g., primary-node (read).

**Root cause:** The group detail API returns { connection_id: number, access_level: string }. The frontend renders connection_id directly as the display label without resolving it to a connection name.

**Steps to reproduce:**

1. Log in as admin.
2. Open Admin → Groups.
3. Grant a connection privilege to any group.
4. Expand the group row.
5. Observe the connection privilege chip shows Connection {id} instead of the connection name.

**Impact:** Minor UX — internal database IDs are exposed to admin users. No security or data integrity risk.

**Fix scope:** Frontend only. The Groups component needs to fetch or receive the connections list and resolve connection_id → name before rendering privilege chips. No API or backend changes required.

**Affected file:** React component rendering the expanded group row (likely AdminGroups.tsx or similar under client/src/).

dpage:
## Summary

Fixes #309: connection-privilege chips in Admin → Groups (expanded row) and Admin → Users showed the internal numeric id, e.g. `Connection 3 (read)`, instead of the connection name like `primary-node (read)`.

`EffectivePermissionsPanel` already resolves the name from a `connections` lookup (falling back to `Connection {id}` only when the id is absent). The real bug was upstream in the fetch: `GET /api/v1/connections` returns a **bare JSON array**, but `AdminGroups.tsx` and `AdminUsers.tsx` decoded it as `{ connections: [...] }` and read `.connections` — always `undefined` for a bare array — so the lookup was empty and every chip hit the fallback. (`AdminTokenScopes.tsx` already handled the bare-array shape, and `ClusterDataContext` decodes it as an array, confirming the shape.)

- Decode the response as an array in both panels so names resolve. One-line change each; `EffectivePermissionsPanel` and `AdminTokenScopes` untouched.

## Scope note

#309 reports the Groups panel; `AdminUsers.tsx` had the **identical** mismatch (its connection chips were silently showing ids too — the same `EffectivePermissionsPanel` is used there), so it's fixed in the same PR for consistency. `AdminTokenScopes.tsx` already coped with the bare array and is left as-is.

## Test plan

- [x] The existing `AdminGroups`/`AdminUsers` tests had mocked `/api/v1/connections` with the **wrong wrapped shape**, which masked the bug; corrected to the real bare-array shape.
- [x] Added regression tests (`AdminGroups.connectionNames.test.tsx`, `AdminUsers.connectionNames.test.tsx`) that render the real `EffectivePermissionsPanel` with a bare-array connections mock, expand a row with `connection_id: 3`, and assert `primary-node (read)` renders while `Connection 3 (read)` does not.
- [x] `npm run lint` clean; affected vitest files pass (84 tests). `AdminGroups.tsx`/`AdminUsers.tsx` ~95% line coverage.

Closes #309

## Summary by CodeRabbit

* **Bug Fixes**
  * Admin Groups and Users panels now display connection names in privilege indicators instead of numeric connection IDs.
  * Updated connection data handling to ensure names resolve correctly in administrative views.

* **Documentation**
  * Added an Unreleased changelog entry describing the corrected connection-privilege display.

### DIFF
--- a/client/src/components/AdminPanel/AdminGroups.tsx
+++ b/client/src/components/AdminPanel/AdminGroups.tsx
@@ -118,10 +118,10 @@ const AdminGroups: React.FC = () => {
     const fetchGroupsAndConnections = useCallback(async (): Promise<RbacGroup[]> => {
         const [groupsData, connResult] = await Promise.all([
             apiGet<{ groups: RbacGroup[] }>('/api/v1/rbac/groups'),
-            apiGet<{ connections?: { id: number; name: string }[] }>('/api/v1/connections').catch(() => null),
+            apiGet<{ id: number; name: string }[]>('/api/v1/connections').catch(() => null),
         ]);
         if (connResult) {
-            setConnections(connResult.connections ?? []);
+            setConnections(connResult ?? []);
         }
         return groupsData.groups || [];
     }, []);
--- a/client/src/components/AdminPanel/AdminUsers.tsx
+++ b/client/src/components/AdminPanel/AdminUsers.tsx
@@ -154,11 +154,11 @@ const AdminUsers: React.FC = () => {
             setError(null);
             const [usersData, connResult] = await Promise.all([
                 apiGet<{ users: RbacUser[] }>('/api/v1/rbac/users'),
-                apiGet<{ connections: { id: number; name: string }[] }>('/api/v1/connections').catch(() => null),
+                apiGet<{ id: number; name: string }[]>('/api/v1/connections').catch(() => null),
             ]);
             setUsers(usersData.users || []);
             if (connResult) {
-                setConnections(connResult.connections || []);
+                setConnections(connResult ?? []);
             }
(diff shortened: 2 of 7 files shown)


moizpgedge (approved):
Verified the root cause against the server code and confirmed the fix live in the UI (Groups and Users panels, including a multi-connection case with non-sequential IDs).
Full test suite passes (3475/3475), lint clean, no regressions — approving.

moizpgedge (approved):
Verified the root cause against the server code and confirmed the fix live in the UI (Groups and Users panels, including a multi-connection case with non-sequential IDs).

PR #332, opened 2026-07-03
fix(server): reject blank connection password on update

### LINKED ISSUE #331: Password storage needs hardening to prevent defense-in-depth gap
When saving a Password in the Server - Connection Details:
  
  The backend's own guard is just "pointer is non-nil" — it doesn't separately reject an explicit empty string. The
  "leave blank to keep unchanged" behavior is currently enforced by the client (which omits the key when blank),
  not by the server independently. If something ever called the API directly with "password": "", the server would
  dutifully encrypt and store an empty password, overwriting the real one. Not something you're likely to hit
  through the UI, but it's a **defense-in-depth gap if you want me to raise it with security-auditor or golang-expert
  for a hardening fix.**

A defense-in-depth gap is a vulnerability in a multi-layered security strategy. It occurs when your defensive layers have an overlapping blind spot, or when the failure of one control compromises the next. This allows attackers to bypass protections, move laterally, or access critical assets.

dpage:
## Summary

- The connection-update datastore guard in `UpdateConnectionFull` only checked that the password pointer was non-nil, so a direct API call with `"password": ""` (or a whitespace-only value) would encrypt and store a blank password, overwriting the real stored credential. The "leave blank to keep unchanged" behaviour was previously enforced only by the web client, leaving a defense-in-depth gap.
- Harden the guard to treat a nil, empty, or whitespace-only password as "keep unchanged", enforced at the datastore layer so no current or future caller can clobber a stored credential with a blank value. The create path, which correctly requires a non-empty password and returns a 400, is unchanged.

## Test plan

- [x] New table-driven integration test `TestUpdateConnectionFullPasswordGuard` covers nil, empty-string, whitespace-only, and non-empty password updates against local Postgres; verified load-bearing (reverting the guard reproduces the overwrite).
- [x] `gofmt -l` clean on changed files; `golangci-lint run ./internal/database/...` reports 0 issues.
- [x] Reviewed by security-auditor: closes the gap with no regression; create/update behaviour is now consistent.

Closes #331

## Summary by CodeRabbit

- **Bug Fixes**
  - Updating a database connection now preserves the saved password when the provided password is empty or whitespace-only.
  - Providing a non-empty password still replaces the stored value as expected.
- **Security**
  - Server-side password updates now follow “leave blank to keep unchanged,” preventing accidental overwrites.
- **Tests**
  - Added end-to-end integration coverage to verify password preservation for `nil`, empty, and whitespace-only inputs, and correct updates for non-empty passwords.
- **Documentation**
  - Updated the changelog in the Security section to reflect the behavior change (referencing issue `#332`).

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -229,6 +229,15 @@ project adheres to
   512M; the collector connection-pool and timeout options are now
   documented in the sample configuration. (#308)
 
+### Security
+
+- Ignore a blank password when updating a database connection, so an
+  empty or whitespace-only password can no longer overwrite the stored
+  credential. The datastore now enforces the "leave blank to keep
+  unchanged" rule server-side as defence-in-depth, rather than relying
+  on the web client alone. The create path, which still requires a
+  non-empty password, is unaffected. (#332)
+
 ## [1.0.0] - 2026-06-08
 
 This release is the first general-availability release of the
--- a/server/src/internal/database/connection_queries.go
+++ b/server/src/internal/database/connection_queries.go
@@ -531,7 +531,16 @@ func (d *Datastore) UpdateConnectionFull(ctx context.Context, id int, params Con
 		args = append(args, *params.Username)
 		argNum++
 	}
-	if params.Password != nil {
+	// On update, a nil, empty, or whitespace-only password means "keep the
+	// existing password unchanged"; we silently skip the password_encrypted
+	// column rather than overwriting it with an encrypted blank value. This
+	// mirrors the documented "leave blank to keep unchanged" client behavior,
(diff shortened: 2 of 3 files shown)


moizpgedge (approved):
Confirmed the vulnerability and fix directly: traced the update handler to show password validation was genuinely missing server-side, then reproduced the exact overwrite bug live via curl against a running instance built from this branch, and again by reverting the one-line guard and rerunning the integration test (both empty-string and whitespace-only cases fail exactly as described, decrypting to the blank value).
Full server test suite passes, gofmt and lint clean, no regressions — approving.

PR #339, opened 2026-07-15
Fix metrics API returning empty response on NaN/Inf samples

susan-pgedge:
## Summary

- The metrics time-series API (`GET /api/v1/metrics/query`) could silently return an empty HTTP 200 response instead of chart data whenever a monitored server's metric sample contained NaN/Infinity (e.g. from a 0/0 division in a system-stats collection function), because Go's JSON encoder cannot marshal non-finite floats and the response status was already written before encoding failed.
- Non-finite samples are now treated the same way the query path already treats missing/NULL data: carried forward from the last known good value, or skipped if none exists yet, so a single bad sample degrades gracefully instead of blanking the entire response.
- `RespondJSON` now logs an encode failure loudly instead of silently swallowing it, so future issues are visible in server logs rather than only manifesting as a mysterious empty UI panel.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes for the full `server` module
- [x] New unit tests added: `TestResolveMetricValue` (NULL gaps and NaN/+Inf/-Inf, both with and without a prior carried-forward value) and `TestRespondJSON_EncodeError` (NaN/+Inf/-Inf)
- [x] `resolveMetricValue` and `RespondJSON` both at 100% function coverage
- [x] `gofmt -l .` clean
- [x] Security review completed (query construction, LOCF staleness risk, log information-disclosure risk) — no blocking findings
- [x] Reproduced the original bug against a live instance (curl against `/api/v1/metrics/query` returned `200` with `Content-Length: 0` for a probe with real NaN samples) before the fix, confirmed real data is returned after

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed metrics time-series queries that returned empty responses when samples contained `NaN` or infinite values.
  * Non-finite samples are now skipped or replaced with the last known valid value.
  * Improved error logging when JSON responses cannot be encoded.

* **Documentation**
  * Added changelog documentation describing the metrics query fix.

This fix was prompted by persistent display of an error message in charts - the message is always there on the Mac, and intermittently on Ubuntu (when not all the servers are hosted on the Ubuntu host):

Screenshot 2026-07-15 at 11 16 56 AM

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -21,6 +21,22 @@ project adheres to
   package matching their chosen embedding provider and model, and
   set `database_path` themselves. (#316)
 
+### Fixed
+
+- Fix the metrics time-series API (`GET /api/v1/metrics/query`)
+  silently returning an empty HTTP 200 response, which the web UI
+  rendered as a generic "No data available" message, whenever a
+  monitored server's metric sample contained a NaN or Infinity value.
+  Go's JSON encoder cannot serialize non-finite floats, so the encode
+  failed silently after the response status had already been sent. The
+  query path now treats a non-finite sample the same way it already
+  treats a missing value from an underlying LEFT JOIN gap; the code
+  carries the last known good value forward when one exists and skips
+  the sample when none does, so a single bad reading degrades
+  gracefully instead of blanking the whole chart. A JSON-encoding
+  failure on this path is now logged loudly rather than silently
+  swallowed, so any future occurrence surfaces in the server logs.
+
 ## [1.0.0] - 2026-06-08
 
 This release is the first general-availability release of the
--- a/server/src/internal/api/response.go
+++ b/server/src/internal/api/response.go
@@ -14,8 +14,8 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"net/http"
-	"os"
 
 	"github.com/pgedge/ai-workbench/server/internal/apiconst"
 )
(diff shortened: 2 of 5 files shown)


dpage (approved):
Reviewed and looks good to merge.

This is a tidy, well-scoped fix for a genuinely nasty failure mode: a single NaN/Inf sample (e.g. a 0/0 division in `system_stats`) caused the JSON encoder to fail *after* the 200 status line had already been flushed, blanking the whole chart with no trace in the logs. Folding non-finite handling into the existing LOCF path that already covered `NULL` gaps is the right call, and `resolveMetricValue` is behaviour-preserving for the previous paths whilst cleanly adding the new one.

A couple of observations, neither blocking:

- The LOCF behaviour now carries the last good value forward across a persistently broken metric, so a stuck probe renders as a flat line rather than an obvious gap. That is consistent with the existing `NULL`-bucket behaviour and is documented, so I would leave it as-is; it is strictly better than before, where the NaN flowed straight through and broke the whole response.
- Swapping `fmt.Fprintf(os.Stderr, ...)` for `log.Printf("[ERROR] ...")` in `RespondJSON` is also a consistency win, bringing it into line with the rest of the `api` package; the logged message carries only the status code and encoder error, so there is no data-leak concern.

Tests are comprehensive (NULL and NaN/±Inf, with and without a carried-forward value, plus the poison-`lastKnown` cases), they pass locally, the changelog entry is accurate, and all CI checks are green.

PR #340, opened 2026-07-15
Fix blank object dashboards: add latest-row metrics query mode

susan-pgedge:
## Symptom

The Table and Index object-detail dashboards in the web client showed blank overview tiles (Live Tuples, Dead Tuples, Table Size, Sequential Scans for tables; equivalent stats for indexes) and a Maintenance Info panel permanently stuck on "Never" for Last Vacuum / Last Autovacuum / Last Analyze / Last Autoanalyze — even against a monitored server with completely normal, healthy activity and confirmed-good underlying data.

## Root cause

Those dashboard panels fetch data by calling `GET /api/v1/metrics/query` with `limit=1&order_by=<column>&order=desc`, expecting a single flat row of real column values back. The metrics API never parsed those three parameters — it silently ignored them and fell through to its normal bucketed time-series behavior, returning `[{name, metric, data: [...]}]` (chart-shaped data) instead. The client's `rows[0]` then picked out a chart-series object instead of a table row, so every field the UI wanted read as `undefined`.

## Fix

- Added a "latest row" query mode (`metrics.QueryLatestRows`, dispatched from `handleMetricsQuery` when `limit`/`order_by` are present) that returns the actual most-recent row(s) of raw probe-table data as flat JSON objects keyed by real column name.
- `order_by` is validated in two layers before it can reach SQL text: a syntactic identifier check, then a semantic check against the probe table's actual discovered columns (or `collected_at`) — only a legitimate, real column name can ever be used, and it's still quoted via the existing `QuoteIdentifier` helper as defense in depth. `order` is allow-listed to exactly `asc`/`desc`. All filter values (connection IDs, database/schema/table/index name) remain fully parameterized.
- Reused the NaN/Infinity-safe float handling from the recent `pg_stat_statements`/`system_stats` dashboard fix so this new path can't reintroduce that class of bug.
- No client changes were needed — the client already sent the correct request and already handled the correct response shape.

## Test plan

- [x] `go build ./...` and `go test ./...` pass for the full `server` module
- [x] New unit tests: valid limit/order_by/order combinations return correctly shaped rows; an `order_by` referencing a non-existent column is rejected with 400 before reaching SQL; an invalid `order` value is rejected with 400; default behavior when `order_by`/`order` are omitted
- [x] New pure/validation logic (`ValidateOrder`, `ResolveOrderByColumn`, `buildLatestRowsQuery`, `sanitizeFloat`) at 100% function coverage; DB-orchestration functions mirror the pre-existing untested-without-a-live-pool condition of `QueryTimeSeries`
- [x] `gofmt` clean
- [x] Security review completed — traced the full `order_by` validation path end to end (including case/whitespace/Unicode/NULL-byte bypass attempts), confirmed RBAC is inherited correctly before this code path is reached, confirmed `limit` is bounded server-side in two places — no blocking findings
- [x] Reproduced the original bug against a live instance (curl with `limit=1&order_by=n_live_tup&order=desc` returned bucketed chart data instead of a row before the fix) and confirmed a real row comes back after

Related: a second, separate bug was found in the same investigation (the Activity Charts on these same dashboards, plus the Estate and Cluster dashboards, request derived/computed metrics like `*_per_sec` and `dead_tuple_ratio` that the metrics API doesn't support) — that will be a follow-up PR, intentionally not included here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Table and Index detail dashboards where overview tiles and the “Maintenance Info” panel could show “Never”.
* **New Features**
  * Added a “latest rows” mode to the metrics API that honors `limit`, `order_by`, and `order`, returning the newest raw values with correct ordering and safe value formatting.
* **Tests**
  * Added unit and database integration coverage for validation, limits/sorting, filtering, and value normalization.
* **Documentation**
  * Updated the unreleased changelog and added guidance for latest-row query behavior.

### DIFF
--- a/.claude/golang-expert/metrics-queries.md
+++ b/.claude/golang-expert/metrics-queries.md
@@ -71,6 +71,57 @@ Metrics tables live under the `metrics` schema
 both read from `connections` by unqualified name; do not add schema
 qualifiers unless the caller's search_path requires it.
 
+## Latest-Row Queries (server)
+
+`server/src/internal/metrics/query.go` exposes a latest-row query path
+alongside the bucketed `QueryTimeSeries`. It returns the most recent raw
+rows of a probe table as flat `map[string]any` objects keyed by real
+column name, so dashboards can read individual dimension and timestamp
+values directly. The public entry point is `QueryLatestRows`; it is
+decomposed into small, separately testable helpers:
+
+- `validateLatestRowParams` validates the probe name, sort direction, and
+  connection list, and clamps the limit to `[1, maxLatestRowLimit]` (100).
+- `discoverLatestRowColumns` confirms the probe exists, discovers its
+  columns via `GetProbeAllColumns`, strips bookkeeping columns with
+  `selectLatestOutputColumns` (`connection_id`, `collected_at`,
+  `inserted_at`), resolves `order_by` against the discovered columns with
+  `ResolveOrderByColumn`, and resolves the database filter column with
+  `ResolveDatabaseColumn` when a `DatabaseName` filter is set.
+- `buildLatestRowsQuery` assembles the SQL; `scanLatestRows` reads rows
+  into flat maps, normalising each value through `normalizeLatestValue`
+  (RFC3339 for `time.Time`, `sanitizeFloat` dropping NaN/Inf to JSON
+  null, `pgtype.Numeric`/`pgtype.Interval` collapsed to seconds/float).
+
+### Identifier safety and the Codacy suppression
+
+Latest-row SQL interpolates only identifiers drawn from a live-discovered
+allow-list: the probe name is checked against
+`information_schema.tables`, the `order_by` column against
+`GetProbeAllColumns`, and every one is `QuoteIdentifier`-wrapped. All
+runtime values bind through `$N` placeholders. golangci-lint/gosec is
+satisfied without a `//nolint`, but Codacy's Opengrep flags the
+`pool.Query(ctx, query, args...)` call in `QueryLatestRows` as
+`go_sql_rule-concat-sqli`. That is a false positive; it is cleared with a
+`// nosemgrep: go_sql_rule-concat-sqli` line immediately above the call,
+kept alongside the existing justification comment. Use `nosemgrep`, not
+`//nolint:gosec`, for Opengrep-only findings.
+
+### Integration tests
+
+DB-executing metrics functions are covered by integration tests in
+`server/src/internal/metrics/query_db_test.go` (package `metrics`, so
+they can exercise unexported helpers). They follow the api package's
+gating convention: skip when `SKIP_DB_TESTS` is set or
+`TEST_AI_WORKBENCH_SERVER` is unset, connect via `pgxpool`, and skip on
+ping failure, so they run in the Server CI jobs and skip cleanly with no
+database. Each test builds its own fixture probe table in the `metrics`
+schema with a representative column mix and drops it on cleanup. Error
+paths that need a failing query (for example the exists-check branch in
+`discoverLatestRowColumns`) are driven with an already-cancelled context;
+the scan-error branch in `scanLatestRows` is driven by passing fewer
+output columns than the query returns.
+
 ## Related Issues
 
 - #56: Alerter FK violations when calculating baselines for deleted
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -37,6 +37,16 @@ project adheres to
   failure on this path is now logged loudly rather than silently
   swallowed, so any future occurrence surfaces in the server logs.
 
+- Fix blank overview tiles and a Maintenance Info panel stuck on
+  "Never" in the Table and Index object-detail dashboards of the
+  web client. These panels requested the single most recent row of
+  stats through the `limit`, `order_by`, and `order` query
+  parameters, but the metrics API ignored those parameters and
+  returned time-bucketed chart data instead. The API now honors
+  these parameters and returns the most recent raw stat rows as
+  real column values, validating `order_by` against the table's
+  known columns before use.
+
 ## [1.0.0] - 2026-06-08
 
 This release is the first general-availability release of the
--- a/server/src/internal/api/metrics_handlers.go
+++ b/server/src/internal/api/metrics_handlers.go
@@ -99,6 +99,16 @@ func (h *MetricsHandler) handleMetricsQuery(
 		return
 	}
 
+	// Latest-row mode: when the caller supplies limit and/or order_by,
+	// return the most recent raw rows instead of a bucketed time series.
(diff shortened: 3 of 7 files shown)


susan-pgedge:
Addressed the review feedback in a follow-up commit (e23016e):

- **order_by allow-list too narrow**: `order_by` is now validated against every column this endpoint actually returns (`outputCols`), not just the numeric metric subset — so `order_by=last_vacuum` or `order_by=relname` are now correctly accepted. The allow-list remains a strict live-discovery check quoted via `QuoteIdentifier`, so this doesn't change the injection-safety property (re-verified with a fresh, focused security pass — no new concerns, safe to ship).
- **`toFloat64` dropped interval Days/Months**: added an `intervalToSeconds` helper that includes Days (86400s) and Months (30-day convention, matching PostgreSQL's own interval-to-scalar arithmetic) alongside Microseconds, for both the `pgtype.Interval` and `*pgtype.Interval` paths. Previously any interval spanning a day or more was silently truncated.
- Added test coverage for both: dimension/timestamp `order_by` acceptance, and interval conversion with non-zero Days/Months (value + pointer + combined cases).

On the Codacy "1 critical security issue" flag: I couldn't retrieve the specific finding (only a category count was exposed via the API), so I traced the full SQL-construction path by hand instead. Every identifier reaching the query text (`probeName`, `orderCol`, output column names) is either a compile-time constant or a value that passed a live `information_schema`-backed allow-list check before being wrapped in `QuoteIdentifier`; all actual data values are parameterized. This matches the identical, never-flagged `fmt.Sprintf`-based SQL-building style already used in this file's pre-existing `BuildMetricsQuery`. My working conclusion is this is a pattern-matching false positive rather than a real finding — happy to revisit if the specific Codacy finding can be surfaced.

All tests pass, `gofmt` clean, new/modified logic at 90%+ coverage.

susan-pgedge:
Followed up on the Codacy findings (commit 438a8136):

**Complexity/length (2 medium)**: `QueryLatestRows` was doing input validation, column discovery, query execution, and row scanning all in one 95-line function (cyclomatic complexity 23, limit 12). Split it into four focused helpers — `validateLatestRowParams`, `selectLatestOutputColumns`, `discoverLatestRowColumns`, and `scanLatestRows` — with `QueryLatestRows` now a ~40-line orchestrator (complexity 4). Pure refactor, no behavior change; all existing tests pass unmodified, new pure helpers at 100% coverage.

**"Critical" SQL Injection (1 critical)**: got the actual finding from the Codacy dashboard — it's flagging line 929, `pool.Query(ctx, query, args...)`, on the pattern "non-literal SQL string reaches a Query call." Traced it fully: the only identifiers interpolated into that string are `probeName`, `orderCol`, and the output column names, each validated against a live-discovered allow-list (`GetProbeAllColumns`, `ResolveOrderByColumn`, the `information_schema.tables` existence check) and then wrapped in `QuoteIdentifier` before reaching the query text. Every actual runtime value (connection IDs, filter strings, limit) is bound through `$N` placeholders in `args`, never concatenated. This confirms the false-positive read from earlier — added a comment at the call site documenting this reasoning for future readers and any automated re-scan.

This finding will likely need a manual "reviewed, not a real issue" dismissal in the Codacy dashboard itself, since I don't have access to do that from here.

susan-pgedge:
Re: the codacy issue, Claude says:

⏺ Clean. Summary:

  - Refactored QueryLatestRows from complexity 23/95 lines down to complexity 4/~40 lines, split into four focused
  helpers, pure behavior-preserving refactor, all tests pass, new pure logic at 100% coverage.
  - Added a WHY comment directly above the pool.Query call at the flagged line, documenting exactly why the SQL
  construction is injection-safe.
  - Confirmed (with the actual finding from your screenshot) that the "critical" flag is a pattern-match false
  positive on line 929 specifically — traced and documented in both the code and a PR comment.
  - Pushed as commit 438a8136, PR comment posted.

  The complexity/length findings should now clear on Codacy's next scan. The security finding will likely keep
  showing up until it's manually marked "reviewed, not a real issue" in Codacy's own dashboard — that requires your
  login, I can't do it from here.

dpage:
I've reviewed this PR and pushed two commits (rebased onto the current `main`, which now includes the NaN/Inf fix from #339) to clear the remaining blockers before it's ready for a human reviewer.

**1. Coverage floor.** The pure validation logic was already at 100%, but the DB-executing functions were largely untested (`QueryLatestRows` 25%, `discoverLatestRowColumns`/`scanLatestRows`/`GetProbeAllColumns` all 0%), well below the project's hard 90% line-coverage floor for new code. I've added `server/src/internal/metrics/query_db_test.go`, a set of DB-backed integration tests that follow the existing gating convention (skip when `SKIP_DB_TESTS` is set or `TEST_AI_WORKBENCH_SERVER` is unset), stand up a synthetic probe-table fixture in the `metrics` schema, and exercise both the happy paths and the defensive error branches. That lifts all four functions above 90% (92.9 / 90.5 / 93.3 / 91.7), verified locally with the race detector against a local Postgres.

**2. Codacy SQL-injection finding.** Codacy's Opengrep flagged the `pool.Query` call in `QueryLatestRows` as `go_sql_rule-concat-sqli`. Having traced the path end to end, I'm satisfied it's a false positive: the only interpolated identifiers are validated against a live-discovered allow-list and `QuoteIdentifier`-wrapped, and every runtime value binds through a `$N` placeholder. I've added a `// nosemgrep: go_sql_rule-concat-sqli` annotation above the call (kept alongside the existing justification comment) so the gate can clear; gosec was already satisfied without a `//nolint`.

The full server suite passes with `-race -p=1`, and both earlier CodeRabbit findings (the `order_by`/`outputCols` allow-list and the interval Days/Months conversion) remain addressed. If Codacy doesn't pick up the inline `nosemgrep` suppression, it may need dismissing as a false positive in the Codacy dashboard.

dpage (approved):
Reviewed in depth, and the two blockers I raised are now resolved and verified.

The latest-row query mode is a clean, well-separated fix, and the SQL-injection surface around the interpolated `order_by` identifier is properly defended in depth: identifier syntax check, then a semantic allow-list against live-discovered columns, then `QuoteIdentifier`, with every runtime value bound through placeholders. RBAC is enforced before the latest-row dispatch. Both earlier CodeRabbit findings (the `order_by`/`outputCols` allow-list and the interval Days/Months conversion) remain addressed.

The two things that were blocking a clean merge have been dealt with: the new DB-executing functions are now covered by integration tests and sit above the 90% floor (92.9 / 90.5 / 93.3 / 91.7), and the Codacy Opengrep SQLi flag, a false positive on quoted-identifier concatenation, is suppressed with `nosemgrep` at both sites. All checks are green, including a full server suite run under `-race -p=1`.

PR #341, opened 2026-07-15
Fix flaky E2E failure from collector schema-migration poll race

susan-pgedge:
## Symptom

The E2E test suite intermittently fails at the "Start stack" step. The server logs:

```
[ERROR] datastore schema health check failed: datastore at postgres@127.0.0.1:5432/postgres:
collector schema is at v2, server requires at least v4; upgrade the collector and restart:
collector schema below minimum required version
```

...and the job then times out waiting for `/health`. This is flaky and unrelated to whatever a given PR actually changes — I observed it fail on the `chromium` browser job while `firefox` and `webkit` passed in the exact same run (each browser job spins up its own independent stack), which is the signature of a race condition rather than a deterministic bug (see [run 29427201083](https://github.com/pgEdge/ai-dba-workbench/actions/runs/29427201083)).

## Root cause

The E2E harness script (`e2e/scripts/apply-collector-schema.sh`) starts the collector in the background so its embedded schema migrations create the operational tables the server needs, then polls the datastore and stops the collector once migrations have landed. The poll condition was:

```bash
if [ -n "${version}" ] && [ "${version}" != "0" ]; then
    echo "==> Collector schema version: ${version}"
    break
fi
```

This declares success as soon as `schema_version` has **any** row — not once the collector has applied **all** of its migrations. The collector applies a sequence of numbered migrations, committing a new `schema_version` row after each one. If the poll happens to land in the window between migration 2 committing and migration 3 starting, the script sees `version=2`, declares victory, and immediately SIGTERMs the collector — killing it mid-migration, before the datastore reaches the version the server's `MinCollectorSchemaVersion` requires.

## Fix

- Added a `--print-latest-schema-version` flag to the collector that reports the newest schema version it knows about (derived from its registered migrations) without touching any database.
- The harness script now asks for that target version up front and waits for `MAX(version)` to actually reach it, instead of stopping at the first non-zero value.
- The target is derived from the collector's own migration registry, so this stays correct automatically as new migrations are added — nothing is hardcoded.
- No change to the collector's actual migration-running logic (`NewDatastore()` etc.) — this is a test-harness fix only.

## Test plan

- [x] `go build ./...`, `go vet ./...`, and `go test ./...` pass for the full `collector` module
- [x] New unit tests: `LatestVersion` (including the empty/zero-value case) and `maybePrintSchemaVersion` (enabled/disabled), both at 100% function coverage
- [x] `gofmt` clean; `bash -n` clean on the modified script
- [x] Verified end-to-end against a live Postgres instance: built the collector, ran the modified script against a fresh throwaway database. It correctly reported "target schema version: 5", waited until the datastore actually reached version 5 (applying all 5 current migrations), then stopped the collector cleanly. Confirmed the old poll condition would have been eligible to stop as early as version 2.
- [x] Re-ran against an already-migrated database to confirm the script remains a fast, correct no-op

## Summary by CodeRabbit

- **New Features**
  - Added a `--print-latest-schema-version` CLI option to output the newest known schema version without starting normal operation.

- **Bug Fixes**
  - Improved end-to-end migration readiness by targeting the exact latest schema version and polling until the datastore reaches it.

- **Tests**
  - Added unit test coverage for computing the latest registered schema version and for the new print flag behavior.

- **Documentation**
  - Updated the changelog to reflect the end-to-end reliability improvement around schema convergence.

### DIFF
--- a/collector/src/database/schema.go
+++ b/collector/src/database/schema.go
@@ -114,6 +114,27 @@ func NewSchemaManager() *SchemaManager {
 	return sm
 }
 
+// LatestVersion returns the highest migration version registered with
+// the schema manager. It reflects the newest schema the collector will
+// converge to once NewDatastore() has applied every pending migration,
+// so callers that need "the version the collector will end up at" can
+// read it without a live database connection or a running migration.
+//
+// The value is derived from the registered migrations, so it stays
+// correct automatically as new migrations are added; nothing needs to
+// track the count separately. It returns 0 when no migrations are
+// registered, which never happens for a SchemaManager built by
+// NewSchemaManager but keeps the helper safe for a zero-value receiver.
+func (sm *SchemaManager) LatestVersion() int {
+	latest := 0
+	for _, m := range sm.migrations {
+		if m.Version > latest {
+			latest = m.Version
+		}
+	}
+	return latest
+}
+
 // registerMigrations registers all available migrations
 func (sm *SchemaManager) registerMigrations() {
 	// Consolidated Migration #1
--- a/collector/src/database/schema_test.go
+++ b/collector/src/database/schema_test.go
@@ -304,29 +304,23 @@ func TestNewSchemaManager(t *testing.T) {
 	}
 }
 
-// TestLatestVersion verifies that LatestVersion reports the highest
-// version among the registered migrations and that it matches the
-// maximum version discovered by scanning the migration slice directly.
-// This keeps the e2e harness's "wait for full convergence" signal
-// correct automatically as migrations are added.
+// TestLatestVersion verifies that LatestVersion computes the maximum
+// registered migration version rather than trusting registration order.
+// The migrations are deliberately supplied out of order (2, 5, 3) so a
+// regression that simply returned the last-registered version would fail
+// this test; only a real max scan yields 5. This keeps the e2e harness's
+// "wait for full convergence" signal correct as migrations are added.
 func TestLatestVersion(t *testing.T) {
-	sm := NewSchemaManager()
-
-	// Derive the expected maximum straight from the registered
-	// migrations so the assertion never needs a hardcoded number.
-	want := 0
-	for _, m := range sm.migrations {
-		if m.Version > want {
-			want = m.Version
-		}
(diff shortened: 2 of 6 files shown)


susan-pgedge:
Addressed both review comments in a follow-up commit (286a5c5e):

- **maybePrintSchemaVersion swallowed its write error**: changed the signature to `(bool, error)` so a failed `fmt.Fprintln` to stdout is no longer silently treated as success. `main()` now exits non-zero with the error on stderr if the write fails, instead of exiting 0 with no output. Confirmed the e2e harness script doesn't need any change — it already treats empty/non-numeric stdout as a failure via its existing numeric-format guard, so this just makes the collector's own exit code honest about what happened.
- **TestLatestVersion couldn't distinguish "max" from "last registered"**: since the real migration registry happens to register in ascending order, the old test would pass even if `LatestVersion()` just returned the last-registered version instead of computing an actual maximum. Rebuilt it around an intentionally unsorted set (2, 5, 3), asserting the result is 5.

All tests pass (including the live-database collector suite), `gofmt` clean, both changed functions at 100% coverage.

dpage (approved):
Reviewed in depth and this is a clean, correctly-diagnosed fix; approving.

The root cause (the harness poll declaring success on the first non-zero `schema_version` row, so a poll landing between two migrations could SIGTERM the collector mid-migration and leave the datastore below the server's required version) is spot on, and the fix is minimal and self-correcting: a database-free `--print-latest-schema-version` flag derives the convergence target from the collector's own migration registry, and the harness waits for `MAX(version)` to reach it, so nothing is hardcoded. It's test-infrastructure only, with no change to the migration-running logic.

I verified it beyond the diff: both new functions (`LatestVersion`, `maybePrintSchemaVersion`) are at 100% coverage, the collector module builds and its DB-backed schema tests pass, `gofmt` and `bash -n` are clean, and running the flag prints a bare integer and exits 0 without touching a database. The CodeRabbit finding on propagating the `fmt.Fprintln` write error was already addressed in 286a5c5. Rebased onto the current main (only the changelog needed reconciling with the two fixes that landed ahead of it).

PR #342, opened 2026-07-16
Fix blank Activity Charts: add per-second rate and dead-tuple-ratio metrics

susan-pgedge:
## Symptom

The "Activity Charts" section of the Table and Index object-detail dashboards ("No tuple operation data available", "No scan data available", "No dead tuple data available") were permanently blank, even against a monitored server with completely normal, healthy activity and confirmed-good underlying data.

## Root cause

Those charts call `GET /api/v1/metrics/query` requesting **computed** metric names that don't exist as literal columns in the underlying `metrics.*` tables:

- Table dashboard: `n_tup_ins_per_sec`, `n_tup_upd_per_sec`, `n_tup_del_per_sec`, `n_tup_hot_upd_per_sec`, `seq_scan_per_sec`, `idx_scan_per_sec`, `dead_tuple_ratio` (probe `pg_stat_all_tables`)
- Index dashboard: `idx_scan_per_sec` (probe `pg_stat_all_indexes`)

The generic time-series endpoint only recognizes literal columns discovered via `information_schema.columns`. Requesting a computed name like `seq_scan_per_sec` (the table only has the raw cumulative counter `seq_scan`) returned `400 metric "seq_scan_per_sec" not found in probe "pg_stat_all_tables"`, which the client's fetch hook rendered as a generic "no data" message.

This same class of computation (rate-per-second from a delta between samples, and a dead-tuple percentage) is already implemented correctly elsewhere in the codebase (`server/src/internal/api/perf_summary_handlers.go`), just never wired into the generic endpoint these two dashboards depend on.

Note: this investigation originally also flagged the Estate and Cluster dashboards as affected by the same pattern (same `_per_sec`/`_ratio` substrings showed up in a grep), but on closer inspection those two components call the existing, already-working `/api/v1/metrics/performance-summary` endpoint, not the generic query path — they were a false-positive match and are unaffected.

## Fix

- Added derived-metric support to the generic query path (`classifyMetrics`, `BuildDerivedMetricsQuery` in `server/src/internal/metrics/query.go`):
  - Any requested metric name ending in `_per_sec` is treated as a per-second rate request. The prefix is validated against the probe's real, discovered numeric columns before use — an unknown base column is rejected with a clear 400, not silently ignored. The rate itself uses the same delta/elapsed-time pattern (with counter-reset and non-positive-elapsed-time guards) already used for the equivalent calculation elsewhere in the codebase.
  - The literal name `dead_tuple_ratio` is accepted only for probes that expose both `n_live_tup` and `n_dead_tup`, computed on the same 0-100 percentage scale the client already expects, using the identical formula used elsewhere in the codebase for the same figure.
  - A single request can mix raw columns and either/both derived-metric kinds.
- Reused/extended the NaN/Infinity-safe float handling so these new computations can't reintroduce that class of bug (this branch predates the separate, still-open NaN-safety fix, so equivalent protection was added independently here).
- No client changes were needed — the client already sent the correct requests and expects exactly this data back.

## Test plan

- [x] `go build ./...` and `go test ./...` pass for the full `server` module
- [x] New unit tests for classification (raw column wins over a coincidental `_per_sec`-suffixed name, valid/invalid `_per_sec` base columns, `dead_tuple_ratio` accepted/rejected by column presence, mixed raw+derived requests preserve order) and for the derived SQL builder (single/multiple rates, ratio scale, filters, all supported aggregations)
- [x] New pure/validation/SQL-building logic at 100% function coverage; DB-orchestration functions mirror the pre-existing untested-without-a-live-pool scoping already accepted for two earlier, related fixes to this file
- [x] `gofmt` clean
- [x] Manually verified both new SQL patterns directly against a live datastore with real table activity (rate and ratio queries both returned correct, sane values)
- [x] Security review completed — traced the full `BaseColumn` validation path (including case/whitespace/cross-probe-column confusion attempts), confirmed `dead_tuple_ratio` only ever uses hardcoded column names, confirmed RBAC is inherited correctly, confirmed no new DoS surface beyond what the existing raw-column path already has — no blocking findings

## Summary by CodeRabbit

* **New Features**
  * Added derived Activity Chart metrics, including per-second rates for counter columns and dead-tuple ratio metrics for table dashboards.
* **Bug Fixes**
  * Fixed Activity Charts that could remain blank on Table and Index object-detail dashboards despite healthy servers.
  * Improved handling of missing/invalid/non-finite metric values to avoid chart rendering failures.
  * Maintained chart continuity across time gaps and sampling intervals.
* **Documentation**
  * Updated the changelog with the Activity Charts fix details.

### DIFF
--- a/.claude/golang-expert/metrics-queries.md
+++ b/.claude/golang-expert/metrics-queries.md
@@ -122,7 +122,72 @@ paths that need a failing query (for example the exists-check branch in
 the scan-error branch in `scanLatestRows` is driven by passing fewer
 output columns than the query returns.
 
+## Derived Metrics in QueryTimeSeries (server)
+
+`QueryTimeSeries` in `server/src/internal/metrics/query.go` serves both raw
+stored columns and computed (derived) metrics from a single request.
+`classifyMetrics` splits the requested metric names, preserving request
+order, into three results: raw column names, a `[]DerivedMetric`, and the
+combined output order. The routing rules are deliberate and order-sensitive:
+
+- A name matching a real numeric column is always a raw metric; a real
+  column wins even when it ends in `_per_sec`.
+- A name ending in `_per_sec` whose prefix is a real numeric column becomes
+  a `DerivedPerSec` rate (delta of the counter over elapsed seconds).
+- The literal `dead_tuple_ratio` is accepted only when the probe exposes
+  both `n_live_tup` and `n_dead_tup`; it is a 0-100 percentage.
+- A repeated name is silently de-duplicated; anything else is a client
+  error.
+
+`BuildDerivedMetricsQuery` builds the derived SQL. It shares the bucketing,
+gap-filling (`generate_series`), filtering, and `$N` argument layout of
+`BuildMetricsQuery`, so the caller scans and applies LOCF identically for
+raw and derived series. Per-second base columns must stay validated against
+the discovered column set and `QuoteIdentifier`-wrapped; never interpolate a
+caller-supplied metric name that has not passed `classifyMetrics`. Negative
+counter deltas (resets/restarts) and non-positive elapsed times are dropped
+to NULL so they never yield a bogus rate.
+
+Both the raw and derived branches feed the shared `scanSeriesRows` helper,
+which scans a bucket-time-plus-N-values result set, applies LOCF per
+connection and metric name via `lastKnown`, and treats a NULL bucket or a
+non-finite sample as a gap.
+
+### NaN/Inf handling: finiteFloat, not resolveMetricValue
+
+Non-finite samples (NaN, +/-Inf) must be treated as gaps, never plotted,
+because `encoding/json` cannot marshal them and one bad value would blank
+the whole response. This guard lives in `toFloat64` via the `finiteFloat`
+helper: `toFloat64` returns `(0, false)` for a non-finite float64/float32/
+`pgtype.Numeric`, and `scanSeriesRows` then treats `!ok` as a gap (LOCF
+carry-forward, or skip when there is no prior value). An earlier design
+(#339) guarded NaN/Inf in a `resolveMetricValue` helper inside the scan
+loop; that was retired in favour of guarding inside `toFloat64`, because the
+`toFloat64` guard is broader and benefits every caller, including the
+latest-row path's `normalizeLatestValue`/`sanitizeFloat`. Do not
+reintroduce `resolveMetricValue`. Note that `toFloat64` still converts
+`pgtype.Interval` through `intervalToSeconds` (handling Days/Months); the
+`finiteFloat` guard and `intervalToSeconds` are independent and both must
+remain.
+
+### Derived-path integration tests
+
+The derived path and `scanSeriesRows` are covered by
+`server/src/internal/metrics/query_timeseries_db_test.go` (same gating
+convention as `query_db_test.go`). Its fixture inserts minute-spaced samples
+with counters rising 60 per minute (a clean 1.0/sec rate) and constant
+live/dead tuple counts (a steady 10% ratio), then exercises raw, `_per_sec`,
+`dead_tuple_ratio`, and mixed requests end-to-end. The two `scanSeriesRows`
+error returns in `QueryTimeSeries` are driven deterministically by passing
+an aggregation that names no SQL function, which makes the built query fail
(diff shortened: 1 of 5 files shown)


susan-pgedge:
Addressed the review feedback in a follow-up commit (c4ff71c2):

- **dead_tuple_ratio ignored the requested aggregation**: the ratio CTE is now aggregation-aware. When `aggregation=last`, it picks the latest sample's `n_live_tup`/`n_dead_tup` per bucket via the same `array_agg(... ORDER BY collected_at DESC))[1]` idiom already used for raw columns and per-second rates, instead of always summing across the bucket. Every other aggregation keeps the prior SUM-based behavior — the emitted SQL for non-`last` requests is byte-identical to before.
- **classifyMetrics didn't deduplicate requested metrics**: a repeated metric name in the request is now silently dropped after its first occurrence, instead of being selected/scanned/emitted twice.
- **Weak test assertion**: the sub-second bucket-width-clamping test now asserts the actual clamped interval arg (`args[0] == "1 seconds"`) instead of only checking that `rate_buckets` appears in the query text — the old version would have passed even if clamping were removed entirely.

Added test coverage for all three: last-vs-other-aggregation ratio semantics, duplicate raw and duplicate derived metric names, and the strengthened clamp assertion. All tests pass, `gofmt` clean, new/modified logic at 100% coverage.

dpage:
Reviewed, and I've rebased this onto the current `main` and closed two gaps before it's ready for a human reviewer.

Since this branch was cut, #339 and #340 both landed on `server/src/internal/metrics/query.go`, so it needed a genuine three-way reconciliation rather than a mechanical rebase:

- **NaN/Infinity handling consolidated.** This PR guards non-finite floats in `toFloat64` via a new `finiteFloat` helper, whilst #339 (now merged) guarded them in a `resolveMetricValue` helper inside the `QueryTimeSeries` bucket loop. This PR's refactor replaces that loop with `scanSeriesRows`, which orphaned `resolveMetricValue`. I kept the `finiteFloat`-in-`toFloat64` approach because it is strictly broader (every caller benefits, including #340's `QueryLatestRows`/`scanLatestRows`), and removed the now-dead `resolveMetricValue` and its test. Behaviour is preserved: `toFloat64` returns not-ok for NaN/Inf, which the LOCF path treats as a gap, exactly as before.
- `toFloat64` retains both #340's `intervalToSeconds` (Days/Months handling) and this PR's finite-float guard.
- All four `### Fixed` changelog entries now coexist.

- **DB-orchestration coverage.** The refactored DB-executing code (`scanSeriesRows` and the derived branch of `QueryTimeSeries`) was uncovered, below the 90% floor. I added `query_timeseries_db_test.go`, DB-backed integration tests (gated on `TEST_AI_WORKBENCH_SERVER` like the existing metrics integration tests) that seed a fixture with rising counters and steady tuple counts and drive `QueryTimeSeries` end to end for raw columns, per-second rates, the dead-tuple ratio, and mixed requests. `scanSeriesRows` is now at 100% and the `QueryTimeSeries` derived path above 90%.

The full server suite passes under `-race -p=1` against a fresh datastore, `golangci-lint` is clean, and #339/#340's tests still pass after the reconciliation. All three original CodeRabbit findings remain addressed.

dpage (approved):
Reviewed in depth, rebased, reconciled, and verified; approving.

The derived-metrics implementation is sound: `_per_sec` base columns are validated against the probe's discovered numeric columns and `QuoteIdentifier`-wrapped, `dead_tuple_ratio` uses only hardcoded tuple columns, and every runtime value stays parameterised, so the SQL-injection surface is properly closed. All three CodeRabbit findings (interval-clamp assertion, applying aggregation to the ratio, deduplicating requested names) are addressed.

I rebased onto the current main and reconciled the three-way overlap with #339/#340: non-finite float handling is now consolidated in `toFloat64` via `finiteFloat` (broader than, and superseding, #339's `resolveMetricValue`, which was removed as dead code once the bucket loop became `scanSeriesRows`), `toFloat64` retains #340's `intervalToSeconds`, and all four changelog entries coexist. I also added DB-backed integration tests covering `scanSeriesRows` (100%) and the `QueryTimeSeries` derived path (>90%), which were previously below the coverage floor.

The full server suite passes under `-race -p=1` against a fresh datastore, `golangci-lint` is clean, and #339/#340's tests still pass after the reconciliation.

PR #343, opened 2026-07-18
Fix missing Table Size and Index Size on object dashboards

susan-pgedge:
## Symptom

The Table and Index object-detail dashboards always showed **Table
Size** and **Index Size** as `--`. Unlike the earlier latest-row
(#340) and derived-metrics (#342) fixes, this wasn't a query-side
gap — no collector probe anywhere gathered relation size, so the
data genuinely never reached the datastore.

## Fix

- `pg_stat_all_tables` probe now collects `table_size` and
  `table_size_pretty` via `pg_table_size`, which covers heap, TOAST,
  and free-space/visibility-map storage while excluding indexes
  (reported separately by the indexes probe). `pg_total_relation_size`
  was deliberately avoided since it would double-count index size,
  and `pg_relation_size` was avoided since it would drop TOAST.
- `pg_stat_all_indexes` probe now collects `index_size` and
  `index_size_pretty` via `pg_relation_size` for each index, across
  both the PG16+ and PG14-15 query variants.
- New collector schema migration, **Version 7**, adds the four
  columns to the `metrics.pg_stat_all_tables` and
  `metrics.pg_stat_all_indexes` partitioned parent tables.
  PostgreSQL 14-18 propagate `ADD COLUMN` on a partitioned parent to
  existing and future partitions automatically, so no per-partition
  DDL is needed. Version 6 was already taken by the separately
  merged embedding-widening migration (#315), so this one correctly
  lands as Version 7.
- No server or client changes were needed. The client's existing
  latest-row request (added in #340) already reads whatever columns
  the probe returns, so the dashboards pick up the new fields with
  no additional plumbing.

## Testing

- New `migration_v7_test.go`: verifies the four columns exist with
  the correct types after migration, that they cascade to a
  partition that already existed before the migration ran, and that
  the migration is idempotent.
- Updated probe unit/integration tests and fixtures for the new
  columns.
- `gofmt -l .` clean, `go build ./...` and `go test ./...` pass for
  the whole collector module.
- Verified live against a locally deployed build: after redeploying,
  `metrics.pg_stat_all_tables.table_size` and
  `metrics.pg_stat_all_indexes.index_size` populate real non-null
  values for monitored tables/indexes, and the Table/Index
  object-detail dashboards render actual sizes instead of `--`.

## Documentation

- `docs/changelog.md`: added an Unreleased/Fixed entry.
- `docs/developer-guide/collector/probe-reference.md`: documented
  the four new columns under the `pg_stat_all_tables` and
  `pg_stat_all_indexes` probe entries.

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL relation size metrics: `table_size` and `index_size` (stored in bytes) for `pg_stat_all_tables` and `pg_stat_all_indexes`, including partition propagation via migration v7.
  * Updated object-detail dashboards to display formatted byte sizes (no longer use “pretty” fields).
* **Bug Fixes**
  * Fixed Table/Index object-detail dashboards where size KPIs could remain stuck as `"--"`.
* **Documentation**
  * Updated the changelog and probe reference to document `table_size`/`index_size` collection.
* **Tests**
  * Added migration and probe coverage for partition propagation and idempotent migration behavior.

### DIFF
--- a/client/src/components/Dashboard/DatabaseDashboard/types.ts
+++ b/client/src/components/Dashboard/DatabaseDashboard/types.ts
@@ -36,7 +36,6 @@ export interface TableLeaderboardRow {
     n_tup_del: number;
     n_tup_hot_upd: number;
     table_size: number;
-    table_size_pretty?: string;
     last_vacuum?: string;
     last_autovacuum?: string;
     last_analyze?: string;
@@ -52,7 +51,6 @@ export interface IndexLeaderboardRow {
     idx_tup_read: number;
     idx_tup_fetch: number;
     index_size: number;
-    index_size_pretty?: string;
 }
 
 /** Leaderboard sort criteria for tables */
--- a/client/src/components/Dashboard/ObjectDashboard/IndexDetail.tsx
+++ b/client/src/components/Dashboard/ObjectDashboard/IndexDetail.tsx
@@ -226,10 +226,9 @@ const IndexDetail: React.FC<ObjectDetailProps> = ({
                 <Box sx={KPI_GRID_SX}>
                     <KpiTile
                         label="Index Size"
-                        value={indexData?.index_size_pretty
-                            ?? formatBytes(
-                                indexData?.index_size ?? null
-                            )}
+                        value={formatBytes(
+                            indexData?.index_size ?? null
+                        )}
                     />
                     <KpiTile
                         label="Index Scans"
--- a/client/src/components/Dashboard/ObjectDashboard/TableDetail.tsx
+++ b/client/src/components/Dashboard/ObjectDashboard/TableDetail.tsx
@@ -347,8 +347,7 @@ const TableDetail: React.FC<ObjectDetailProps> = ({
                     />
                     <KpiTile
                         label="Table Size"
-                        value={tableData?.table_size_pretty
-                            ?? formatBytes(tableSize)}
+                        value={formatBytes(tableSize)}
                     />
                     <KpiTile
                         label="Sequential Scans"
--- a/client/src/components/Dashboard/ObjectDashboard/__tests__/IndexDetail.test.tsx
+++ b/client/src/components/Dashboard/ObjectDashboard/__tests__/IndexDetail.test.tsx
@@ -1,0 +1,297 @@
+/*-------------------------------------------------------------------------
+ *
+ * pgEdge AI DBA Workbench
+ *
+ * Copyright (c) 2025 - 2026, pgEdge, Inc.
+ * This software is released under The PostgreSQL License
+ *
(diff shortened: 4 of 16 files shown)


susan-pgedge:
## Codacy SQL-injection finding: confirmed false positive, suppressed

Codacy flagged a CRITICAL "SQL Injection" issue at
`collector/src/database/migration_v7_test.go:140`:

```go
if _, err := pool.Exec(ctx, ddl); err != nil {
```

This is a false positive. The `ddl` string (built via `fmt.Sprintf`
just above) is constructed entirely from:

- Hardcoded test constants declared as `const` in the test
  (`partitionStart = "2026-05-11"`, `partitionEnd = "2026-05-18"`,
  `partitionSuffix = "20260511"`).
- `childIdent` / `parentIdent`, which are produced by pgx's own
  `pgx.Identifier{...}.Sanitize()` safe identifier-quoting function.

There is no user input or untrusted external data anywhere in the
construction. Codacy's static pattern-matcher flags any
`pool.Exec(ctx, <non-literal-string-variable>)` regardless of
provable safety.

This is the same class of false positive already resolved on PR #340
in `server/src/internal/metrics/query.go`, using a
`// nosemgrep: go_sql_rule-concat-sqli` suppression comment plus a
WHY comment. This commit applies the same treatment: a matching
`nosemgrep` suppression and an explanatory comment above the call.

Note: this adds the standard `nosemgrep` suppression in source. I have
not independently confirmed whether Codacy's own gate clears purely
from the inline suppression, or whether it also requires a human to
mark the finding as a false positive in the Codacy dashboard. If the
check does not go green automatically, it may need that manual
dashboard action by someone with Codacy access.

susan-pgedge:
Addressed two items in commit 41ccf3ad:

**Doc wording accuracy (CodeRabbit finding):** Updated both `docs/developer-guide/collector/probe-reference.md` and `docs/changelog.md` to note that `pg_table_size` also covers visibility-map storage, alongside heap, TOAST, and free-space-map storage. The descriptions now document the complete `pg_table_size` footprint (indexes still excluded, as the index probe reports those separately).

**Migration idempotency test fix:** Hardened the Version 7 collector schema migration idempotency test in `collector/src/database/migration_v7_test.go`. Verified by temporarily breaking migration idempotency and confirming the test then fails as expected, before reverting the deliberate break.

dpage (approved):
Revised to collect only raw byte counts (`table_size`/`index_size`); the redundant `pg_size_pretty` columns are dropped and the client formats bytes with its existing `formatBytes` helper. Full CI matrix green, coverage floor met on touched code, CodeRabbit clean.

PR #344, opened 2026-07-18
Fix Estate dashboard panels getting stuck on "No data" after a transient failure

susan-pgedge:
## Symptom

After a brief backend restart or any transient API failure, the
Estate Overview dashboard's KPI tiles (XID Age, Cache Hit Ratio,
Transactions, Checkpoints) and Event Timeline could get stuck
showing "No data" indefinitely. The only way to recover was to
manually click the sidebar's Refresh button or reload the page —
nothing retried on its own.

## Root cause

This client has no React Query or SWR. Every panel hook only
refetches when a single shared `lastRefresh` timestamp
(`ClusterDataContext`) changes, and that timestamp only advances
when the sidebar's `GET /api/v1/clusters` call succeeds. If an
individual panel's own fetch failed transiently, or if the shared
refresh call itself failed, nothing retried automatically and the
affected panel just sat blank.

## Fix

- New shared `useRetryingFetch` hook (`client/src/hooks/useRetryingFetch.ts`)
  providing capped exponential backoff: 3s, 6s, 12s, 24s, capped at
  45s. On success it stops retrying; a manual refresh (change in
  `resetKey`/`lastRefresh`) always pre-empts a pending retry.
- Wired into:
  - `useTimelineEvents` (Event Timeline)
  - `usePerformanceSummary` (XID Age / Cache Hit Ratio / Transactions
    / Checkpoints tiles)
  - `KpiTilesSection` (Estate KPI tiles) — also adds a subtle
    "Reconnecting..." indicator while a retry is pending, so a
    transient outage reads differently from genuinely empty data
  - `StatusPanel`'s `fetchAlertsData`

## Testing

- New unit tests for `useRetryingFetch` and the modified hooks;
  updated existing suites for the new return shapes.
- `npm run lint`: 0 errors (39 pre-existing warnings, none in
  touched files).
- Full suite: 164 test files / 3386 tests pass.
- Touched-file line coverage: `useRetryingFetch.ts` 98.46%,
  `useTimelineEvents.ts` 90.66%, `usePerformanceSummary.ts` 98.3%,
  `KpiTilesSection.tsx` 100%. `StatusPanel/index.tsx` is a large
  pre-existing file at ~71% overall coverage; every line touched by
  this change is covered, and the residual gap is pre-existing,
  untouched code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Dashboard KPI tiles, performance summaries, alerts, and timeline events now automatically retry after temporary API failures using capped exponential backoff.
  - The UI no longer gets stuck on “No data” after brief backend interruptions.
  - KPI tiles and related panels show **“Reconnecting…”** while a retry is pending, and resume normal rendering after recovery.
  - Manual refresh now pre-empts pending retries to immediately re-fetch fresh data.
- **Documentation**
  - Updated the changelog to reflect the improved dashboard recovery behavior.

### DIFF
--- a/client/src/components/Dashboard/EstateDashboard/KpiTilesSection.tsx
+++ b/client/src/components/Dashboard/EstateDashboard/KpiTilesSection.tsx
@@ -118,6 +118,11 @@ const KpiTilesSection: React.FC<KpiTilesSectionProps> = ({ selection, serverIds 
 
             const alertCount = (alertsData.alerts || []).length;
 
+            // A late resolution after unmount must not touch state; a
+            // superseded/unmounted attempt is not a failure, so report
+            // success to keep it out of the retry schedule.
+            if (!isMountedRef.current) { return true; }
+
             setAggregate({
                 totalServers,
                 totalConnections,
--- a/client/src/components/Dashboard/EstateDashboard/KpiTilesSection.tsx
+++ b/client/src/components/Dashboard/EstateDashboard/KpiTilesSection.tsx
@@ -21,6 +21,7 @@ import { formatNumber } from '../../../utils/formatters';
 import { KPI_GRID_SX } from '../styles';
 import { countEstateServers } from '../../../utils/clusterHelpers';
 import { logger } from '../../../utils/logger';
+import { useRetryingFetch } from '../../../hooks/useRetryingFetch';
 import type { EstateSelection } from '../../../types/selection';
 
 interface KpiTilesSectionProps {
@@ -62,12 +63,16 @@ const KpiTilesSection: React.FC<KpiTilesSectionProps> = ({ selection, serverIds 
     const [error, setError] = useState<string | null>(null);
     const isMountedRef = useRef<boolean>(true);
     const initialLoadDoneRef = useRef<boolean>(false);
+    const { run, retrying } = useRetryingFetch({
+        resetKey: lastRefresh,
+        enabled: !!user && serverIds.length > 0,
+    });
 
     const totalServers = useMemo(() => countEstateServers(selection), [selection]);
     const serverIdsKey = serverIds.join(',');
 
-    const fetchAggregateData = useCallback(async (): Promise<void> => {
-        if (!user || serverIds.length === 0) { return; }
+    const fetchAggregateData = useCallback(async (): Promise<boolean> => {
+        if (!user || serverIds.length === 0) { return true; }
 
         if (!initialLoadDoneRef.current) {
             setLoading(true);
@@ -84,7 +89,7 @@ const KpiTilesSection: React.FC<KpiTilesSectionProps> = ({ selection, serverIds 
                 ),
             ]);
 
-            if (!isMountedRef.current) { return; }
+            if (!isMountedRef.current) { return true; }
 
             let totalConnections = 0;
             let transactionRate = 0;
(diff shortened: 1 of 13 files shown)


susan-pgedge:
Addressed the four CodeRabbit findings in 148da3f2:

**1. Missing license header** — Added `This software is released under The PostgreSQL License` to the copyright header of `client/src/components/StatusPanel/__tests__/StatusPanel.refetch.test.tsx`.

**2. Stale in-flight attempts in `useRetryingFetch.ts`** — Added a monotonic attempt-generation counter (`generationRef`). Each `execute()` captures the generation current at its start; `run()`, `cancel()`, and the `resetKey` effect all increment it. When a fetch promise resolves, its captured generation is compared against the current one — if superseded, the result is ignored entirely (no `failureCountRef` bump, no `retrying` state change, no retry timer). Added two regression tests reproducing the race via both a fresh `run()` and a `resetKey` change while an older attempt is still pending, then resolving that stale attempt with a failure and asserting `retrying` stays false and no extra fetch fires.

**3. Alert-refresh call sites bypassing retry handling** — Introduced a shared retry-managed `refetchAlerts` callback wrapping `runAlertsFetch(fetchAlertsData)` in `client/src/components/StatusPanel/index.tsx`, and replaced every direct `fetchAlertsData()` call — both acknowledgement-refresh handlers and the unacknowledge reconciliation — with it (the mount/refresh effect now routes through it too). Added `StatusPanel.acknowledge.test.tsx` covering both the single and group ack-confirm refresh paths.

**4. `KpiTilesSection` treating failed responses as success** — The fetch now returns `false` immediately when either the performance-summary or alerts response is non-OK, before constructing any aggregate/partial data, so the retry controller sees the failure. Updated the existing test to assert the `Reconnecting…` indicator appears on a non-OK response and that a subsequent successful retry recovers.

Verification: `npm run lint` (0 errors), full suite `165 files / 3390 tests` passing, and touched-file line coverage above the 90% floor (useRetryingFetch.ts 97.3%, KpiTilesSection.tsx 96.8%, and the modified `StatusPanel/index.tsx` lines all covered).

dpage:
Pushed a follow-up commit (`Strengthen retry tests and tidy consumer fetch guards`) addressing the review, then rebased onto the latest `main`:

- **Acknowledge tests** now reject the first post-ack alerts fetch, advance the base retry delay, and assert the scheduled retry lands and the alert re-renders. Verified they genuinely fail if the ack refresh bypasses the retry controller (call count 1 instead of 2), so they now prove the managed path rather than merely that a follow-up fetch occurred. This closes the last open CodeRabbit thread.
- **`useTimelineEvents`** gains a retry-heal test for parity with the other three consumers.
- Added a consistent `isMountedRef` guard before the terminal `setState` in `KpiTilesSection`, `usePerformanceSummary`, and `StatusPanel.fetchAlertsData` (the last had none), all returning `true` so the success/failure contract is preserved.
- Stabilised `connectionIds` via a sorted-joined key in `useTimelineEvents`, matching the existing `eventTypesKey` treatment.

Full CI matrix green; touched lines clear the 90% coverage floor. The core `useRetryingFetch` hook was reviewed and left unchanged.

dpage (approved):
Retry-with-backoff resilience is well built: the generation-counter design correctly handles the stale-in-flight-attempt race, manual refresh pre-empts pending retries, and non-OK responses drive the backoff. Review follow-ups (stronger acknowledge/timeline retry tests, consistent unmount guards, connectionIds stabilisation) are in; CI green and CodeRabbit clean.

Issue #294, opened 2026-06-09
Memory embedding fails with non-OpenAI providers due to hardcoded vector(1536) schema


imtiazqa:
The chat_memories table is created with vector(1536) which matches OpenAI's embedding dimensions. When Gemini or Voyage embedding providers are configured, store_memory fails with a dimension mismatch error.

Gemini (gemini-embedding-001, gemini-embedding-2) generates 3072-dimension vectors:

Failed to store memory: failed to insert memory: ERROR: expected 1536 dimensions, not 3072 (SQLSTATE 22000)
Voyage (voyage-3, voyage-3-lite, voyage-2) generates 1024 or 512-dimension vectors:

Failed to store memory: failed to insert memory: ERROR: expected 1536 dimensions, not 1024 (SQLSTATE 22000)
Steps to Reproduce:

- Configure the server with Gemini or Voyage as the embedding provider.
- Ask Ellie to store a memory.
- Tool fails with dimension mismatch error.
- Expected: Memory stores successfully regardless of embedding provider.

Root Cause:

The schema hardcodes vector(1536) instead of using a dimension that matches the configured embedding provider. Additionally, pgvector HNSW and IVFFlat indexes both enforce a 2000-dimension limit for the vector type, so widening to vector(3072) for Gemini breaks index creation.

Workaround (pgvector ≥ 0.7.0):

For Gemini (3072 dimensions) — use halfvec which supports HNSW up to 4000 dims:

DROP INDEX IF EXISTS idx_chat_memories_embedding;

ALTER TABLE chat_memories
ALTER COLUMN embedding TYPE halfvec(3072)
USING embedding::halfvec(3072);

CREATE INDEX idx_chat_memories_embedding
ON chat_memories USING hnsw (embedding halfvec_cosine_ops);
Verified working on pgvector 0.8.1.

For Voyage (1024 dimensions) — standard vector type works since 1024 is within pgvector's index limit:

DROP INDEX IF EXISTS idx_chat_memories_embedding;

ALTER TABLE chat_memories
ALTER COLUMN embedding TYPE vector(1024)
USING embedding::vector(1024);

CREATE INDEX idx_chat_memories_embedding
ON chat_memories USING hnsw (embedding vector_cosine_ops);
Proper Fix:

Make the vector column type and dimension configurable based on embedding_provider at schema initialization, or cap Gemini output at 1536 dimensions using the output_dimensionality API parameter.

imtiazqa:
[...]

PR #352, opened 2026-07-21
Fix latest-row metrics query returning stale historical data

susan-pgedge:
## Symptom

The Table and Index object-detail dashboards' overview tiles and
Maintenance Info panel could show stale, pre-maintenance values
indefinitely, even after the collector had already recorded fresh,
correct data.

Reproduced live: running `ANALYZE` on `northwind.public.orders`
corrected its live-tuple-count estimate downward (1933 → 1367), but
the dashboard kept showing the old, larger tuple count and a stale
(`Never`) Last Analyze value indefinitely afterward, despite the
datastore holding the correct fresh sample.

## Root cause

The latest-row mode of `GET /api/v1/metrics/query` (`limit` +
`order_by`, added by #340) built:

```sql
ORDER BY <order_by column> <asc|desc>, collected_at DESC
LIMIT <limit>
```

`order_by` was the *primary* sort key across the probe table's
**entire history** for the matching rows, with `collected_at DESC`
only breaking exact ties. So a request for "the latest row of table
X, sorted by `n_live_tup`" didn't return the most recent sample — it
returned whichever historical sample had the numerically highest
`n_live_tup`, which is exactly wrong for any column that can decrease
over time: tuple counts after `ANALYZE`/`VACUUM`/`DELETE`, dead-tuple
counts after `VACUUM`, and so on.

## Fix

`buildLatestRowsQuery` now always reduces to each monitored entity's
(table's or index's) newest sample by `collected_at` first, via an
inner `DISTINCT ON` over the entity-key (schema/table or
schema/index) columns, and only then ranks those already-latest rows
by the requested `order_by` column. Probes with no entity-key columns
keep a flat `ORDER BY ... collected_at DESC`.

A request filtered down to one specific table or index — which is
what `TableDetail.tsx`/`IndexDetail.tsx` always do — now always
returns that entity's true latest sample, regardless of which column
it asks to sort by. Confirmed no real caller today relies on
multi-entity ranking through this endpoint (the leaderboard widgets
use a separate, already-`DISTINCT ON`-correct `/api/v1/metrics/latest`
endpoint), but the fix keeps that case correct too, since it groups
by entity before ranking.

All identifiers remain validated against a live-discovered allow-list
and `QuoteIdentifier`-wrapped; all values stay `$N`-bound — no change
to the existing SQL-injection-safe pattern.

## Testing

- Rewrote the integration fixture with three distinct entities, one
  sampled so its newest value is smaller than an older historical
  peak — the exact regression.
- New/updated unit and integration tests confirm a single-entity
  query now returns the newest row (not the historical max) and a
  multi-entity query ranks by each entity's own latest sample.
- `gofmt -l .` clean, `go build ./...` clean.
- Full server suite passes; integration tests verified live against a
  real PostgreSQL instance.
- Touched-function line coverage: `buildLatestRowsQuery` 100%,
  `latestEntityKeyColumns` 100%, `discoverLatestRowColumns` 90.5%,
  `QueryLatestRows` 91.7%, `scanLatestRows` 93.3%.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Table and Index object-detail dashboards showing stale overview tiles and Maintenance Info after fresh data.
  * Latest metrics now consistently return each entity’s newest `collected_at` sample (including correct behavior when ordering fields decrease over time), with proper single- and multi-entity ranking and `limit`.
  * Dimension columns no longer include `*_pretty` value columns, preventing entities from being split into multiple identity groups.
* **Documentation**
  * Updated the unreleased changelog entry describing the dashboard fix.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -148,6 +148,25 @@ project adheres to
   coexist, and the effective access for each connection is the
   higher of the two. (#302)
 
+- Fix the Table and Index object-detail dashboards showing stale,
+  pre-maintenance values in their overview tiles and Maintenance Info
+  panel indefinitely, even after the collector recorded fresh, correct
+  data. Running ANALYZE on a table corrected its live-tuple-count
+  estimate downward, yet the dashboard kept showing the old, larger
+  count and a stale or missing Last Analyze value. The latest-row mode
+  of the metrics API (`GET /api/v1/metrics/query` with `limit` and
+  `order_by`) ranked the table's entire history by the requested
+  `order_by` column and used `collected_at` only to break exact ties,
+  so it returned whichever historical sample had the highest column
+  value rather than the most recent sample. Any column that can
+  decrease over time, such as a tuple count after ANALYZE, VACUUM, or
+  DELETE, or a dead-tuple count after VACUUM, could therefore return a
+  permanently stale historical row. The query now reduces to each
+  monitored entity's newest sample by `collected_at` first, then ranks
+  those already-latest rows by the requested `order_by` column, so a
+  request filtered to one table or index always returns that entity's
+  true latest sample regardless of the sort column.
+
 ## [1.0.0] - 2026-06-08
 
 This release is the first general-availability release of the
--- a/server/src/internal/api/latest_handlers.go
+++ b/server/src/internal/api/latest_handlers.go
@@ -390,23 +390,16 @@ func resolveDatabaseColumnCached(
 }
 
 // BuildDimensionColumns identifies entity key columns suitable for the
-// DISTINCT ON clause. These are text/name-type columns that are not
-// internal bookkeeping columns.
+// DISTINCT ON clause. It delegates to metrics.EntityKeyColumns so this
+// endpoint and the generic latest-row query path share one definition
+// of what counts as an identity column and can never drift apart. The
+// shared helper excludes text-typed value columns (e.g. the "_pretty"
+// size renderings) that would otherwise fragment a single entity.
 func BuildDimensionColumns(
 	allColumns []string,
 	colTypes map[string]string,
 ) []string {
-	var dims []string
-	for _, col := range allColumns {
-		if internalColumns[col] {
-			continue
-		}
-		dt := colTypes[col]
-		if dt == "text" || dt == "character varying" || dt == "name" {
-			dims = append(dims, col)
-		}
-	}
-	return dims
+	return metrics.EntityKeyColumns(allColumns, colTypes)
 }
 
 // queryLatestSnapshot builds and executes the DISTINCT ON query that
--- a/server/src/internal/api/latest_handlers_test.go
+++ b/server/src/internal/api/latest_handlers_test.go
@@ -151,6 +151,32 @@ func TestBuildDimensionColumns(t *testing.T) {
 			},
 			want: []string{"datname"},
 		},
+		{
+			// Simulates PR #343 adding table_size / table_size_pretty to
+			// the leaderboard table. The text-typed *_pretty rendering is
+			// a value, not identity; including it in DISTINCT ON would
+			// fragment one table across its changing rendered sizes and
+			// silently corrupt the leaderboard ranking. It must be
+			// excluded even though it is text.
+			name: "text _pretty value column excluded from dimensions",
+			allColumns: []string{
+				"connection_id", "collected_at", "inserted_at",
+				"database_name", "schemaname", "relname",
+				"table_size", "table_size_pretty", "n_live_tup",
+			},
+			colTypes: map[string]string{
+				"connection_id":     "integer",
+				"collected_at":      "timestamp with time zone",
+				"inserted_at":       "timestamp with time zone",
+				"database_name":     "text",
+				"schemaname":        "name",
+				"relname":           "name",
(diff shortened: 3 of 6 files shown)


susan-pgedge:
Pushed an additional commit (642a793d) after live-testing this fix locally combined with the not-yet-merged PR #343 (adds `table_size`/`table_size_pretty` to `pg_stat_all_tables`, `index_size`/`index_size_pretty` to `pg_stat_all_indexes`).

**Regression found:** the original entity-key heuristic here (`latestEntityKeyColumns`) classified *any* text/varchar/name-typed column as part of a monitored entity's identity. `table_size_pretty` is a text-typed **value** column (e.g. `"576 kB"`), not an identity column, but it got swept into the `DISTINCT ON` grouping anyway. Since it changes over time (`NULL` before #343's migration ran, populated after), the query fragmented one real table into multiple "entities" and reproduced the exact staleness bug this PR fixes — confirmed live via `curl` against a locally combined build of both branches.

Same flawed heuristic also already exists in shipped code: `BuildDimensionColumns` in `server/src/internal/api/latest_handlers.go`, which backs the Table/Index Leaderboard's `/api/v1/metrics/latest` endpoint. It hasn't broken in production yet only because `table_size_pretty`/`index_size_pretty` don't exist on `main` yet — but it would as soon as #343 merges.

**Fix:** added a shared `metrics.IsEntityKeyColumn`/`EntityKeyColumns` that excludes any column whose name ends in `_pretty` (this codebase's established convention for a `pg_size_pretty`-rendered companion of another raw value column) in addition to internal bookkeeping columns. Verified no genuine identity column anywhere in the schema ends in `_pretty`, so this is a pure bug fix with no behavior change for any probe on current `main`. `BuildDimensionColumns` now delegates to the same shared helper so both call sites can't drift apart again.

Re-verified live end-to-end after this fix: Table Overview tile and Maintenance Info panel for `northwind.public.orders` now correctly show the post-ANALYZE live-tuple count (1,367) and populated Last Analyze timestamp, instead of the stale pre-ANALYZE row.

susan-pgedge:
Addressed the CodeRabbit feedback in commit `00add067`:

**Major/functional — missing `connection_id` in entity key:**
`buildLatestRowsQuery` now always includes `connection_id` in the
`DISTINCT ON` entity key, in addition to any text/name dimension
columns. Without it, if a caller ever requested multiple
`connection_id`s in one query, two monitored servers sharing a
schema+table name would collapse into a single `DISTINCT ON` group and
one connection's latest sample would be silently dropped.

**Fallback simplification:**
Because `connection_id` is present on every probe table, the entity key
is never empty, so the old no-entity-key fallback branch (which ordered
by `orderCol` then `collected_at` directly — the original staleness
pattern) is now redundant and has been removed. Every probe uses the
same `DISTINCT ON (connection_id[, keys]) ... ORDER BY ...,
collected_at DESC` wrapping. I verified all existing unit and live-DB
integration cases still hold before removing the branch.

**Test gap — fallback didn't exercise a real metric column:**
Added `TestQueryLatestRows_NoEntityKeyIntegration`, a live-DB test using
a probe with no text/name entity keys. Its fixture has three samples
where the newest (`cpu_user` 50) is *not* the historical maximum
(`cpu_user` 99); with `order_by=cpu_user desc, limit=1` it asserts the
query returns 50 (newest by `collected_at`), catching the exact
regression this PR fixes. The string-level unit test was also updated
to rank by a real metric column instead of `collected_at`.

**Lint failure:**
Fixed the `misspell` failure in `query_test.go` ("analogue" -> "analog")
that was blocking CI.

**Changelog blank line:**
Verified — the blank line between the new bullet and the preceding
`(#294)` bullet is already present in the committed changelog (line
100), so no change was needed there.

`gofmt -l`, `go build ./...`, and `make lint` (golangci-lint) all pass
clean; the metrics package tests pass against a live DB, and all touched
functions remain at or above the 90% coverage floor
(`buildLatestRowsQuery` at 100%).

dpage:
Reviewed with a Go correctness pass and a dedicated security pass (dynamic SQL construction). Both came back clean: the `DISTINCT ON (connection_id[, keys])` reduction correctly returns each entity's newest sample, the duplicate-column concern doesn't materialise because `selectLatestOutputColumns` strips `connection_id`/`collected_at`/`inserted_at` (verified against live PG), and the SQL-injection-safe invariant holds since the new `distinctClause` is fed solely from schema introspection and `QuoteIdentifier`-wrapped.

Also confirmed the cross-PR interaction with #343: this was built against #343's original form (with `*_pretty` columns), but #343 merged without them, so the `_pretty` exclusion is now a harmless general guard and nothing in the tests/fixtures breaks on rebase.

Pushed one small follow-up commit documenting the `outputCols` precondition on `buildLatestRowsQuery`, then rebased onto the latest `main`, resolving the `docs/changelog.md` conflict by keeping all three Unreleased/Fixed bullets. Full CI matrix (server on PG 14-18, E2E, docker builds, Codacy, CodeRabbit) is green.

dpage (approved):
Correct, well-tested fix for the historical-peak-vs-newest bug; the DISTINCT-ON-by-collected_at-first reduction is sound and the connection_id entity-key inclusion is the right call. Go correctness and security reviews both clean; CI green across PG 14-18.

PR #353, opened 2026-07-21
Fix Index Scan Activity chart always showing no data

susan-pgedge:
## Symptom

The Index object-detail dashboard's Scan Activity chart always
showed "No index scan data available" — for every index, with no
exceptions.

## Root cause

Two compounding gaps, confirmed live via `curl`:

1. **Client:** `IndexDetail.tsx`'s Scan Activity chart query sent the
   index's own name (e.g. `pk_orders`) as the `table_name` parameter:
   ```js
   tableName: objectName,  // objectName = "pk_orders"
   ```
   The server filters `table_name` on `relname` (the table name), so
   `table_name=pk_orders` matched zero rows — no table is ever named
   after an index. Confirmed: `table_name=pk_orders` → `"data":[]`;
   `table_name=orders` (the real table) → real data, but that would
   wrongly blend every index on the table together.

2. **Server:** even a corrected client request would have needed a
   server-side fix. `metricQueryBase` in
   `server/src/internal/metrics/query.go` — shared by the raw-column
   and derived-metric time-series query builders that back Activity
   Charts — only filtered on `DatabaseName`/`SchemaName`/`TableName`.
   It had no `IndexName` handling at all, unlike the separate
   latest-row query path (added by #340), which already supports
   `index_name`. The client also had no `indexName` field anywhere to
   send one with (`useMetrics.ts`/`MetricQueryParams` only knew about
   `schemaName`/`tableName`).

## Fix

- `metricQueryBase` now filters on `indexrelname` when `IndexName` is
  set, mirroring the existing `SchemaName`/`TableName` pattern
  exactly (`$N`-bound, no probe-column-existence check — consistent
  with how those two already behave). Purely additive: omitting
  `index_name` produces byte-identical SQL to before, so the Table
  Detail dashboard's existing Activity Charts are unaffected.
- `handleMetricsQuery` (time-series mode) now parses `index_name`,
  mirroring the latest-row handler's existing parsing.
- Client: added `indexName` to `MetricQueryParams`, wired it into
  `useMetrics.ts`'s query-string builder, and `IndexDetail.tsx`'s
  Scan Activity chart now sends `indexName: objectName` instead of
  the incorrect `tableName: objectName` (and sends no `tableName`, so
  it scopes to the specific index rather than the whole table).

## Testing

- Server: new unit tests proving `index_name` binds to
  `indexrelname = $N` for both the raw-column and derived-metric
  paths, and that omitting it leaves the generated SQL/args
  unchanged. New live-DB integration tests confirm a matching index
  returns real rate data and a non-matching one returns empty, for
  both query paths. Handler test added for the new parsing.
- Client: new/updated tests confirm the Scan Activity chart request
  now carries `index_name=<index name>` and no `table_name`; existing
  `TableDetail.tsx` Activity Charts are confirmed unaffected (they
  never set `indexName`, so `index_name` is never appended).
- `gofmt`/`go vet`/`golangci-lint`: clean. Full Go suite (unit + live
  integration) passes. `npm run lint`: 0 errors. Full client suite:
  3364 tests pass.
- Touched-function coverage: server `metricQueryBase`,
  `BuildMetricsQuery`, `BuildDerivedMetricsQuery` all 100%; client
  `useMetrics.ts` 100% lines, `IndexDetail.tsx` 98.14% lines.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the Index dashboard’s **Scan Activity** chart to correctly scope metrics to the selected index.
  * Prevented scan activity data from mixing across different indexes by ensuring metrics requests use the correct index-scoping filter.
  * Improved handling of index-scoped metric queries so the chart returns data reliably (and not “No index scan data available” incorrectly).
* **Documentation**
  * Updated the changelog with details about the **Scan Activity** chart fix and index filtering improvements.

### DIFF
--- a/client/src/components/Dashboard/ObjectDashboard/IndexDetail.tsx
+++ b/client/src/components/Dashboard/ObjectDashboard/IndexDetail.tsx
@@ -145,7 +145,7 @@ const IndexDetail: React.FC<ObjectDetailProps> = ({
         connectionId,
         databaseName,
         schemaName,
-        tableName: objectName,
+        indexName: objectName,
         timeRange: timeRange.range,
         buckets: CHART_BUCKETS,
         aggregation: 'avg',
--- a/client/src/components/Dashboard/ObjectDashboard/__tests__/IndexDetail.test.tsx
+++ b/client/src/components/Dashboard/ObjectDashboard/__tests__/IndexDetail.test.tsx
@@ -15,7 +15,7 @@ import { ThemeProvider, createTheme } from '@mui/material/styles';
 import IndexDetail from '../IndexDetail';
 import type { IndexDetailData } from '../types';
 import type { UseMetricsReturn } from '../../../../hooks/useMetrics';
-import type { MetricSeries } from '../../types';
+import type { MetricSeries, MetricQueryParams } from '../../types';
 
 // ---------------------------------------------------------------------------
 // Mocks
@@ -47,8 +47,15 @@ let mockMetricsReturn: UseMetricsReturn = {
     error: null,
     refetch: vi.fn(),
 };
+// Spy that captures the params IndexDetail hands to useMetrics for the
+// Scan Activity chart, so the routing tests can assert the query is scoped
+// to the index (indexName) rather than misrouted into tableName.
+const mockUseMetrics = vi.fn();
 vi.mock('../../../../hooks/useMetrics', () => ({
-    useMetrics: () => mockMetricsReturn,
+    useMetrics: (params: MetricQueryParams | null) => {
+        mockUseMetrics(params);
+        return mockMetricsReturn;
+    },
 }));
 
 vi.mock('../../../../contexts/useAICapabilities', () => ({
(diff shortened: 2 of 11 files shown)


susan-pgedge:
Addressed CodeRabbit's feedback on `index_name` handling with two fixes (commit `69addb4e`):

**Server test quality (`server/src/internal/api/metrics_handlers_test.go`)** — CodeRabbit correctly flagged that `TestHandleMetricsQuery_TimeSeriesMode_ParsesIndexNameFilter` used an invalid `buckets=9999` value, which made the handler return a 400 *before* the code path that consumes `filters.IndexName` (i.e. before `QueryTimeSeries` ran). The test therefore passed even if `index_name` parsing were removed. I restructured it to use only valid parameters and to inject a fake query function via a new seam on `MetricsHandler` (`queryTimeSeriesFn`, defaulting to `metrics.QueryTimeSeries` when nil — the same narrow function-field pattern already used elsewhere in the `api` package, e.g. `visibilityListerFn`). The test now asserts the parsed `IndexName` actually reaches the query layer; a mutation of the parse makes it fail. A companion integration test (`...DefaultQueryFn`) covers the nil-fallback default path so the seam wiring stays exercised.

**Client (`client/src/hooks/useMetrics.ts`)** — the initial-load reset effect did not depend on `params.indexName`, so switching indexes on the Index detail dashboard did not refetch the Scan Activity chart. Added `params?.indexName` to that effect's dependency array, with hook-test coverage.

Verification: `gofmt -l` clean, `go build ./...` and `go vet ./...` pass; full server suite passes against a live DB (`-p 1`), stable across repeated runs; all modified server lines are covered per `go tool cover`; client `useMetrics` tests pass (20/20).

dpage:
Reviewed with a Go correctness pass and a React pass. Both clean: the `indexrelname = $N` filter is injection-safe (hardcoded column, bound value) with provably correct `argNum` sequencing across all filter combinations and both query paths, and the client refetches on index switch via the `params` object-reference mechanism (verified by a test that would fail if the swap were reverted).

Rebased onto the latest `main` (which gained #343 and #352 today) and resolved the resulting conflicts: the changelog kept all bullets; `query.go`/`query_test.go`/`IndexDetail.tsx` auto-merged and were verified semantically correct (both #352's latest-row DISTINCT-ON work and this PR's time-series filter coexist without interference); and the add/add `IndexDetail.test.tsx` conflict was reconciled by keeping #343's full component suite and grafting this PR's two index-routing guard tests (dropping duplicates, including one that relied on #343's removed `index_size_pretty`).

Also folded in two small review follow-ups: the `useMetrics` reset effect now watches `tableName`/`schemaName` too (coherent loading-spinner behaviour), and a comment documenting why `metricQueryBase`'s last filter omits `argNum++`.

Full CI matrix green (server PG 14-18, client, E2E, docker, Codacy, CodeRabbit); both PRs' live-DB integration tests pass; touched code clears the 90% floor.

dpage (approved):
Correct, minimal, well-tested full-stack fix. Server argNum sequencing verified across all filter combinations and injection-safe; client refetch-on-index-switch verified. Rebased onto main with #343/#352 conflicts resolved and both merges verified by the respective experts; review follow-ups (S1/S2) folded in. CI green across the matrix.

Issue #336, opened 2026-07-09
Cache hit ratio graph not rendering properly on the SERVER dashboard


susan-pgedge:
This issue should be fixed in PR 359: https://github.com/pgEdge/ai-dba-workbench/pull/359

I've added the Northwind sample database on a server; even after a refresh and restart, the cache hit ratio graph is not rendering properly for the new database, though it does render properly when I select the database:

<img width="1015" height="649" alt="Image" src="https://github.com/user-attachments/assets/600e52f2-3a62-4669-b4d1-b4c700194fbd" />

<img width="711" height="309" alt="Image" src="https://github.com/user-attachments/assets/340a9b17-d36e-4d6c-975e-6be513fc6928" />

susan-pgedge:
Claude says:

⏺ Confirmed the root cause directly in the code — this is an ECharts axis-scaling edge case, not what the first
  pass suspected.

  File: client/src/components/Chart/options/common.ts:166-173 (buildYAxis())

  export function buildYAxis(): object {
      return {
          type: 'value',
          axisLabel: { formatter: formatNumericValue },
      };
  }

  No min/max is set, so ECharts auto-computes the axis range from the data. When every point in a series has the
  same value, ECharts collapses min === max, giving a zero-height range — the line/area has nothing to interpolate
  against, so it renders nothing.

  That's exactly northwind's situation: it's a brand-new, essentially read-only sample database, so its
  cache-hit-ratio time series has been a flat, unbroken 100.0 the entire time. The other databases (99.9%) have
  tiny natural fluctuation across samples, so ECharts computes a real (if narrow) range and the area fill renders
  normally.

susan-pgedge:
This issue should be fixed by PR 359.

PR #361, opened 2026-07-22
Fix blank Database Summaries chart for databases with no history yet

susan-pgedge:
## Symptom

Reported while investigating #336/#359: on a Server dashboard's
"Database Summaries" section, a database could show a correct
current Cache Hit Ratio percentage (e.g. 100.0%) right next to a
completely blank chart box, indistinguishable from a rendering
glitch.

## Root cause

`current` and `time_series` for each card come from two independent
server queries in `server/src/internal/api/perf_summary_handlers.go`:
`queryDatabaseStats` (latest single row) and
`queryDatabaseCacheHitTimeSeries` (24h bucketed history). A database
with only a few minutes of monitoring history can have a valid
`current` value while the bucketed query legitimately returns zero
rows, so the API returns `time_series: []` alongside a valid
`current`.

On the client, `DatabaseSummariesSection.tsx` guarded the chart with
`db.cache_hit_ratio?.time_series && (...)`. An empty array is truthy
in JavaScript, so the chart `Box` still rendered. Inside it,
`Sparkline` explicitly returns `null` for zero-length data, leaving an
empty box next to a populated percentage.

This is unrelated to #336/#359 (an ECharts axis-scaling min===max
collapse for flat-value series) — a different failure mode in a
different component. `Sparkline.tsx` and its existing null-for-empty
test are untouched.

## Fix

`DatabaseSummariesSection.tsx` now checks
`time_series.length > 0` before rendering the `Sparkline`, and shows a
small "No history yet" placeholder (matching the muted-secondary-text
empty-state convention used elsewhere in this file/directory) in the
same-sized chart area otherwise.

## Test plan

- [x] New `DatabaseSummariesSection.test.tsx`: populated time_series
      still renders the Sparkline; empty time_series with valid
      current shows the placeholder; undefined cache_hit_ratio shows
      the placeholder; plus existing loading/error/empty-state
      coverage.
- [x] `npx eslint` clean on both touched files
- [x] `npx vitest run` — 37/37 passing across ServerDashboard + the
      untouched Sparkline test
- [x] Coverage of the modified file: 98.5% lines, above the 90% floor

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the Cache Hit Ratio card to display “No history yet” when cache-hit history is missing or empty, instead of showing a blank chart.
  * Kept the current Cache Hit Ratio percentage visible, with consistent placeholders when related metrics are unavailable.
* **Tests**
  * Added a comprehensive test suite for loading, error handling, empty states, chart visibility, metric formatting, and keyboard interactions for database card overlays.

### DIFF
--- a/client/src/components/Dashboard/ServerDashboard/DatabaseSummariesSection.tsx
+++ b/client/src/components/Dashboard/ServerDashboard/DatabaseSummariesSection.tsx
@@ -278,7 +278,8 @@ const DatabaseSummariesSection: React.FC<ServerSectionProps> = ({
                                 </Typography>
                             </Box>
 
-                            {db.cache_hit_ratio?.time_series && (
+                            {db.cache_hit_ratio?.time_series
+                                && db.cache_hit_ratio.time_series.length > 0 ? (
                                 <Box sx={{ height: 30, mt: 0.5, mb: 5 }}>
                                     <Sparkline
                                         data={toSparklineData(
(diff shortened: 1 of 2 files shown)


dpage (approved):
Reviewed via /review-pr. Clean, well-scoped fix for the blank Cache Hit Ratio chart when a database has no bucketed history yet: the empty-array-is-truthy guard bug is correctly diagnosed and fixed, with thorough tests (loading/error/empty/populated/keyboard paths) and coverage above the project floor. Follow-up commit extracts the shared chart-box dimensions into a single CHART_BOX_SX constant so the sparkline and placeholder cannot drift out of lockstep. Lint clean, all client tests pass, and full CI is green.

Issue #360, opened 2026-07-22
Database Summaries card renders blank chart for databases with insufficient history


susan-pgedge:
## Symptom

On a Server dashboard's "Database Summaries" section, a database can
show a correct current Cache Hit Ratio percentage (e.g. 100.0%) next
to a completely blank chart area, while other database cards at the
same percentage render a normal filled sparkline. Observed with
newly-added databases that have only a few minutes of monitoring
history.

## Root cause

`current` and `time_series` for each database card come from two
independent server queries in
`server/src/internal/api/perf_summary_handlers.go`:

- `current` (`queryDatabaseStats`): reads the single most-recent
  `pg_stat_database` row.
- `time_series` (`queryDatabaseCacheHitTimeSeries`): buckets rows
  over the requested time range (24h) via `date_bin`, grouped by
  `datname, bucket`.

A database with only a few minutes of collection history can have a
valid single-row `current` value while the bucketed time-series query
returns zero rows, so the API legitimately returns `time_series: []`
alongside a valid `current`.

On the client, `DatabaseSummariesSection.tsx` guards the chart with:

```tsx
{db.cache_hit_ratio?.time_series && (
    <Box sx={{ height: 30, mt: 0.5, mb: 5 }}>
        <Sparkline
            data={toSparklineData(db.cache_hit_ratio.time_series)}
            color={getCacheColor(db.cache_hit_ratio?.current ?? 0)}
            height={30}
        />
    </Box>
)}
```

An empty array `[]` is truthy in JavaScript, so the `Box` still
renders. Inside it, `Sparkline.tsx` explicitly returns `null` when
`data.length === 0`, leaving an empty box next to a populated
percentage — not an ECharts axis-scaling bug (unrelated to #336),
just a missing empty-state for "insufficient history yet."

## Proposed fix

When `time_series` is empty but `current` is valid, render a small
"No history yet" placeholder in place of the blank chart area,
instead of silently rendering nothing.

PR #363, opened 2026-07-22
Fix ghost database cards persisting after a database is dropped

susan-pgedge:
## Symptom

Reported while investigating #360: a database dropped from a
monitored server continues to appear as a card in the Server
dashboard's "Database Summaries" section for up to the full selected
time range (24h by default, or 7d/30d on those tabs), even though it
no longer exists and the collector already recorded its absence.

## Root cause

`handleDatabaseSummaries` (`server/src/internal/api/perf_summary_handlers.go:671`)
builds its result as a union of five independent queries sharing one
`dbMap`. `queryDatabaseSizes`, `queryDatabaseStats`, and
`queryDeadTupleRatios` are restricted to the latest collected
snapshot and self-correct once a database is dropped (the next poll
simply excludes it). But `queryDatabaseCacheHitTimeSeries` scans the
**entire requested time range**, so a database with any historical
samples inside that window — including right up until the moment it
was dropped — kept re-creating a `dbMap` entry, resurrecting a ghost
card long after every other signal said it no longer existed.

Reproduced live: dropped two real databases, confirmed the
collector's own `metrics.pg_database` latest snapshot correctly
stopped listing them within a few poll cycles, yet their Database
Summaries cards persisted.

## Fix

`queryDatabaseSizes` (reading `metrics.pg_database`, PostgreSQL's own
catalog and the source of truth for what currently exists) is now the
only function permitted to create `dbMap` entries. The other four
query functions (`queryDatabaseStats`, `queryDeadTupleRatios`,
`queryTransactionRates`, `queryDatabaseCacheHitTimeSeries`) now only
enrich an existing entry and skip rows for databases absent from that
base set, instead of creating one of their own.

Side effect (desirable): template databases (`template0`/`template1`)
are now consistently excluded from every path, since only
`queryDatabaseSizes`'s `datistemplate = false` filter can introduce an
entry.

## Test plan

- [x] New `perf_summary_database_summaries_test.go`: seeds a `keep`
      database (present in every latest snapshot) alongside a `ghost`
      database that only has historical `pg_stat_database` rows inside
      the query window — confirms `ghost` and `template0` are excluded
      while `keep` gets a fully populated summary (no regression).
      Also covers: enrichment queries never create entries against an
      empty/foreign base set; query errors (dropped tables) are
      logged and skipped without panicking; scan errors (`NULL` in
      NOT NULL columns) are skipped without creating spurious entries.
- [x] `gofmt` clean, `go vet` clean, `golangci-lint` clean
- [x] Full `internal/api` package tests pass
- [x] Coverage of touched functions: 100% on
      `queryDatabaseSizes`/`queryDatabaseStats`/`queryDeadTupleRatios`/
      `queryDatabaseCacheHitTimeSeries`; 82.4% on
      `queryTransactionRates` — the only gap is a per-row scan-error
      branch that's unreachable by construction for this query (the
      self-join guarantees a non-null `datname`, and the `CASE...ELSE 0`
      guarantees a non-null `tx_rate`); left in place for parity with
      its sibling functions rather than removed to chase the number.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed obsolete database entries from performance summaries when historical metrics reference databases no longer present.
  * Improved database metric enrichment to ignore unknown databases.
  * Added safeguards for missing metrics data and invalid database records, preventing incomplete or erroneous summary entries.

### DIFF
--- a/server/src/internal/api/perf_summary_database_summaries_test.go
+++ b/server/src/internal/api/perf_summary_database_summaries_test.go
@@ -1,0 +1,443 @@
+/*-------------------------------------------------------------------------
+ *
+ * pgEdge AI DBA Workbench
+ *
+ * Copyright (c) 2025 - 2026, pgEdge, Inc.
+ * This software is released under The PostgreSQL License
+ *
+ *-------------------------------------------------------------------------
+ */
+package api
+
+import (
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+	"github.com/pgedge/ai-workbench/server/internal/database"
+)
+
+// databaseSummariesTestSchema mirrors the minimum columns the five
+// database-summaries query helpers read. It lives in the metrics schema
+// because the queries fully qualify their table names with metrics.*.
+const databaseSummariesTestSchema = `
+CREATE SCHEMA IF NOT EXISTS metrics;
+DROP TABLE IF EXISTS metrics.pg_database CASCADE;
+DROP TABLE IF EXISTS metrics.pg_stat_database CASCADE;
+DROP TABLE IF EXISTS metrics.pg_stat_all_tables CASCADE;
+
+CREATE TABLE metrics.pg_database (
+    connection_id        integer     NOT NULL,
+    collected_at         timestamptz NOT NULL,
+    datname              text        NOT NULL,
+    datistemplate        boolean     NOT NULL DEFAULT false,
+    database_size_bytes  bigint
+);
+
+CREATE TABLE metrics.pg_stat_database (
+    connection_id  integer     NOT NULL,
+    collected_at   timestamptz NOT NULL,
+    datname        text,
+    numbackends    integer     NOT NULL DEFAULT 0,
+    blks_hit       bigint      NOT NULL DEFAULT 0,
+    blks_read      bigint      NOT NULL DEFAULT 0,
+    xact_commit    bigint      NOT NULL DEFAULT 0,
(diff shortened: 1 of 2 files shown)


dpage (approved):
Reviewed via /review-pr, with a deeper Go pass from a domain review.

The enrich-only guard is the correct fix for the ghost-card bug: making `queryDatabaseSizes` (latest `metrics.pg_database` snapshot, `datistemplate = false`) the sole creator of `dbMap` entries stops `queryDatabaseCacheHitTimeSeries`'s whole-window scan from resurrecting dropped databases, and desirably excludes template databases too. Call ordering is guaranteed (sizes is Query 1), the four new DB-backed tests genuinely pinpoint the buggy path rather than passing trivially, connection-leak hygiene in the error test is handled deliberately, and I ran the suite locally against 127.0.0.1: 4/4 pass. gofmt/vet/lint clean; CodeRabbit and Codacy clean.

I initially flagged a possible availability regression (a transient `pg_database` failure wiping all cards), but on closer analysis it does not exist: all five queries share one read-only transaction, so a genuine query error poisons the transaction and takes down all five equally, both before and after this change (the PR's own `TestDatabaseSummaries_QueryError` documents this by using a fresh transaction per helper). The only way sizes yields an empty map whilst the others have data is when there genuinely are no non-template user databases, where an empty result is correct. No mitigation warranted.

Minor notes, all acceptable as-is: `queryTransactionRates` sits at 82.4% because its self-join makes the per-row scan-error branch unreachable (documented in the PR); a NULL `database_size_bytes` would drop a card, but the collector's `pg_database_size()` probe errors rather than storing NULL, so it can't happen in practice. LGTM.

PR #365, opened 2026-07-23
Fix "Hide monitoring queries" missing Workbench-internal traffic

susan-pgedge:
## Symptom

The Server dashboard's "Top Queries" section has a "Hide monitoring
queries" toggle. With it on, Workbench-internal overhead queries
against the app's own `ai_workbench` metadata database still showed
up, e.g. `INSERT INTO "metrics"."pg_stat_s...`, `WITH
recent_statements AS (...`, `WITH freeze_settings AS (...`.

## Root cause

`handleTopQueries`'s `exclude_collector` filter only excluded queries
containing the collector's `ai_dba_wb_probe` marker literal, which
`collector/src/probes/base.go` wraps only around the collector's
read-only probe SELECTs against monitored databases. Since
`ai_workbench` shares its Postgres instance with monitored databases,
`pg_stat_statements` captures instance-wide activity, including two
unmarked categories of Workbench's own traffic:

1. The collector's own batched `INSERT INTO metrics.*` writes that
   persist collected metrics — storage writes, never wrapped in the
   probe marker.
2. The alerter's (and server's) own internal reads of that same
   metrics schema for alert evaluation — e.g. the `recent_statements`
   and `freeze_settings` CTEs in
   `alerter/src/internal/database/metric_registry.go`, an entirely
   separate binary the collector's marker has no bearing on.

## Fix

The exclusion clause now also matches the `metrics` schema qualifier
combined with the `pg_`/`spock_` table-name prefixes every internal
metrics table uses by convention, in both quoted (`pgx.Identifier`,
the collector's writes) and unquoted (hand-written SQL, the
alerter's/server's reads) forms.

Matching on the table-name prefix rather than the bare `metrics`
schema name is a deliberate choice: it avoids hiding a genuinely
monitored external database that happens to expose its own
application schema also named `metrics`. `pg_` is reserved by
PostgreSQL and `spock_` is the Spock extension's namespace, so a
collision with real user tables is effectively impossible, and any
future internal metrics table following the same convention is
covered automatically.

Query construction is extracted into a `buildTopQueriesQuery` helper
so the SQL is directly testable in isolation from the HTTP/RBAC
plumbing, matching the pattern already used by sibling handlers in
this file.

## Test plan

- [x] `TestTopQueries_Issue364_ExcludeCollector` — probe-marked,
      quoted collector metrics writes, and unquoted alerter/server
      metrics reads are all excluded; a normal user query and external
      queries against a database-level `metrics` schema (including one
      shaped like `metrics.pgx_data`, guarding the underscore-escape)
      are preserved.
- [x] `TestTopQueries_Issue364_NoExcludeReturnsAll` — toggle off
      returns everything, unchanged.
- [x] `TestTopQueries_Issue364_HandlerHTTP` — full handler over HTTP
      confirms the same filtering end-to-end.
- [x] `TestBuildTopQueriesQuery_ClauseSelection` plus handler
      validation/error-path tests (query error, scan error, begin-tx
      error, param validation).
- [x] `gofmt`/`go vet`/`golangci-lint` clean
- [x] Coverage: `buildTopQueriesQuery` 100%, `handleTopQueries` 95.4%
      (previously untested)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the top-queries performance summary filter to reliably exclude Workbench’s internal monitoring activity.
  * Preserved legitimate queries, including those involving similarly named schemas and tables.
  * Improved handling of invalid requests and database errors with safer, predictable responses.

* **Tests**
  * Added comprehensive coverage for filtering, validation, query limits, ordering, and error scenarios.

### DIFF
--- a/server/src/internal/api/perf_summary_handlers.go
+++ b/server/src/internal/api/perf_summary_handlers.go
@@ -148,6 +148,110 @@ var validTopQueryOrderColumns = map[string]bool{
 	"shared_blks_read": true,
 }
 
+// workbenchInternalExclusionClause filters pg_stat_statements rows that
+// represent AI DBA Workbench's own internal traffic rather than genuine
+// activity on the monitored database. Because the ai_workbench metadata
+// database shares its PostgreSQL instance with monitored databases,
+// instance-wide pg_stat_statements captures Workbench's own overhead. The
+// clause removes three categories of that internal traffic:
+//
+//  1. Collector probe SELECTs, which the collector wraps in a marker
+//     column literal (ai_dba_wb_probe) before running them against the
+//     monitored database (see collector/src/probes/base.go).
+//  2. The collector's own batched writes into the internal metrics schema,
+//     emitted as INSERT INTO "metrics"."<table>" ... via pgx.Identifier,
+//     which quotes both the schema and table identifiers.
+//  3. The alerter's and server's own reads of the internal metrics schema,
+//     hand-authored as unquoted metrics.<table> references (for example
+//     the alerter's slow_query_count and age_percent CTEs in
+//     alerter/src/internal/database/metric_registry.go).
+//
+// Categories 2 and 3 match the "metrics" schema qualifier combined with
+// the pg_ / spock_ table-name prefixes that every internal metrics table
+// uses by convention, in both the unquoted (hand-written SQL) and quoted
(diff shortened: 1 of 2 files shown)


susan-pgedge:
Combined into #367, which now targets `main` directly and contains this PR's full change plus the #366 follow-up. Closing this one so there's a single PR to merge, with no ordering dependency.

Issue #364, opened 2026-07-23
"Hide monitoring queries" toggle doesn't hide most Workbench-internal query overhead


susan-pgedge:
## Symptom

On the Server dashboard's "Top Queries" section, toggling "Hide
monitoring queries" on still shows Workbench-internal overhead
queries against the app's own `ai_workbench` metadata database, e.g.:

- `INSERT INTO "metrics"."pg_stat_statements" (...)`
- `WITH recent_statements AS ( SELECT ... )`
- `WITH freeze_settings AS ( SELECT ... )`

## Root cause

`handleTopQueries` (`server/src/internal/api/perf_summary_handlers.go:1113`)
implements the filter as:

```go
excludeCollectorClause = "AND pss.query NOT LIKE '%ai_dba_wb_probe%'"
```

The `ai_dba_wb_probe` marker (`collector/src/probes/base.go:32`,
`collector/src/probes/change_tracking.go:26`) is wrapped only around
the collector's read-only probe SELECTs run against monitored
databases:

```go
"SELECT '%s' AS ai_dba_wb_probe, subq.* FROM (%s) AS subq"
```

Since `ai_workbench` (the app's own metadata database) lives on the
same Postgres instance as monitored databases, `pg_stat_statements`
captures instance-wide activity, including Workbench's own internal
traffic against its own database. Two categories of that traffic are
never marked and slip through the filter:

1. The collector's own bulk `INSERT INTO metrics.*` statements that
   persist collected metrics into the datastore. These are storage
   writes, not probes, so `base.go`'s wrapper never touches them.
2. The **alerter**'s internal alert-evaluation queries against its own
   `ai_workbench` metrics store, e.g. the `recent_statements` and
   `freeze_settings` CTEs in
   `alerter/src/internal/database/metric_registry.go:888` and
   `:1045`. The alerter is a separate binary; the collector's marker
   mechanism has no bearing on it at all.

## Proposed fix

Broaden the exclusion beyond the single probe marker to also catch
the collector's metrics-store writes and the alerter's metrics-store
reads. A schema-qualified pattern match against the internal
`metrics` schema (covering both read and write access to
`metrics.*` tables) would catch all three categories uniformly,
though the exact implementation needs care to avoid false-positives
if a genuinely monitored database happens to have its own schema
also named `metrics`.

PR #367, opened 2026-07-23
Fix "Hide monitoring queries" showing Workbench-internal traffic

susan-pgedge:
## Symptom

Follow-up to #364/#365. With "Hide monitoring queries" on, Top Queries
still showed only `ai_workbench` (the app's own maintenance database)
rows — confirmed live after exercising a real monitored `northwind`
database with a genuine workload, none of its activity ever surfaced.

## Root cause

#365's `exclude_collector` filter matched query *text* against known
schema/table patterns. That approach is fundamentally incomplete —
found two more categories live by querying the raw metrics store:

1. `SELECT column_name, data_type FROM information_schema.columns
   WHERE table_schema = $2 AND table_name = $1` — a schema-
   introspection query where the schema/table names are bound
   parameters, not literal text. No text pattern can ever match this.
2. `WITH data_buckets AS (SELECT ... FROM metrics."pg_stat_statements"
   ...)` — a third quoting style (schema unquoted, table quoted)
   #365's patterns didn't anticipate.

Because the maintenance database's own collector-driven traffic
accumulates far more total execution time than a freshly exercised
monitored database, these unfiltered rows dominated a
`total_exec_time`-sorted top-10 list and crowded out genuine
monitored-database activity entirely.

## Fix

Replace text-pattern matching with a database-identity check:
`buildTopQueriesQuery` now excludes every row whose resolved
`database_name` matches the server's configured maintenance database
(`database.database` in the YAML config, e.g. `ai_workbench` — read
via a new `maintenanceDBName` parameter threaded through
`NewPerfSummaryHandler`, not hardcoded). This catches all of the
maintenance database's own traffic categorically, regardless of
query-text shape, quoting, or parameterization.

The `ai_dba_wb_probe` marker clause is kept — it still matters for
collector probe SELECTs that run against *other* monitored databases,
which the database-identity check can't catch since those rows are
attributed to the monitored database, not the maintenance one. The
now-redundant `metrics.pg_*`/`metrics.spock_*` text patterns are
removed, since the database-identity check supersedes them entirely.

## Test plan

- [x] Fixtures cover both previously-unfilterable cases (parameterized
      `information_schema` query, mixed-quoting `metrics."pg_stat_statements"`
      CTE) — both now excluded by identity, not text.
- [x] A marker-tagged probe attributed to a monitored (`northwind`)
      database is still excluded via the surviving marker clause.
- [x] A genuine `northwind` row whose text happens to superficially
      resemble `metrics."pg_stat_statements"` is preserved — proving
      the filter now goes by identity, not text (would have been
      wrongly dropped under #365's approach).
- [x] `TestBuildTopQueriesQuery_ClauseSelection` rewritten to assert
      clause presence/absence, parameter binding order, and that
      obsolete text fragments are gone.
- [x] `gofmt`/`go vet`/`golangci-lint` clean; full `internal/api` and
      `cmd/mcp-server` suites pass
- [x] Coverage: `buildTopQueriesQuery` 100%, `NewPerfSummaryHandler`
      100%, `handleTopQueries` 95.4%

## Summary by CodeRabbit

* **New Features**
  * Improved the Top Queries performance summary to optionally exclude internal maintenance and collector-generated activity.
  * Added more accurate filtering while preserving genuine monitored-database queries.
  * Enhanced handling of query limits, filtering, and validation.

* **Bug Fixes**
  * Improved resilience when performance data cannot be queried or processed, returning safer results instead of failing unexpectedly.

* **Tests**
  * Added comprehensive coverage for filtering, validation, error handling, and HTTP behavior.

### DIFF
--- a/server/src/cmd/mcp-server/handlers.go
+++ b/server/src/cmd/mcp-server/handlers.go
@@ -195,8 +195,16 @@ func SetupHandlers(deps *HandlerDependencies) func(*http.ServeMux) error {
 		timelineHandler := api.NewTimelineHandler(deps.Datastore, deps.AuthStore, rbacChecker)
 		registerDatastoreHandler(mux, timelineHandler, authWrapper, "Timeline events", deps.Datastore)
 
-		// Performance summary endpoint (for performance dashboard)
-		perfHandler := api.NewPerfSummaryHandler(deps.Datastore, deps.AuthStore)
+		// Performance summary endpoint (for performance dashboard).
+		// Pass the configured maintenance (datastore) database name so the
+		// top-queries endpoint can filter the Workbench's own traffic when
+		// exclude_collector is set (issue #366).
+		maintenanceDBName := ""
+		if deps.Config != nil && deps.Config.Database != nil {
+			maintenanceDBName = deps.Config.Database.Database
+		}
+		perfHandler := api.NewPerfSummaryHandler(
+			deps.Datastore, deps.AuthStore, maintenanceDBName)
 		registerDatastoreHandler(mux, perfHandler, authWrapper, "Performance summary", deps.Datastore)
 
 		// Metrics query endpoints (for monitoring dashboards)
--- a/server/src/internal/api/perf_summary_database_summaries_test.go
+++ b/server/src/internal/api/perf_summary_database_summaries_test.go
@@ -96,7 +96,7 @@ func newDatabaseSummariesTestHandler(
 	}
 
 	ds := database.NewTestDatastore(pool)
-	handler := NewPerfSummaryHandler(ds, nil)
+	handler := NewPerfSummaryHandler(ds, nil, "")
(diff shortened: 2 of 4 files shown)


susan-pgedge:
@coderabbitai review

dpage:
I don't think this is the correct way to handle this - it would exclude *all* queries on the workbench's data store, not just the workbench's own queries. This is undesirable as users might have other tools or services that are making use of the data that's collected, that needs monitoring.

susan-pgedge:
If I try to selectively weed out the queries that are not on the ai_workbench database to only those not involved in queries, my actual exercising of my sample database doesn't show up.

Would the right answer be to add a toggle/way to select the database you want to see the queries from?

susan-pgedge:
Closing for now - we may want to add a database-scoped query tile later.

PR #372, opened 2026-07-27
fix(client): normalise tool schemas to the library's wire contract

### LINKED ISSUE #370: Ask Ellie: tool calls fail with "tools.0.custom.input_schema: Input does not match the expected shape"
## Summary

Every request to `POST /api/v1/llm/chat` (and `/chat/stream`) that includes a `tools` array fails with:

```
anthropic (400): tools.0.custom.input_schema: Input does not match the expected shape.
```

This is 100% reproducible with any single well-formed MCP tool — it is not about a specific tool's schema content, it's systemic. In practice this breaks essentially all tool-using chat, including Ask Ellie for any question that requires a tool call.

## Root cause

Two parts of the current API disagree on field casing for a tool's JSON Schema:

- `GET /api/v1/mcp/tools` (the tool-discovery endpoint the client calls) returns each tool's schema under the camelCase key `inputSchema` — see the documented `MCPTool` schema in `server/src/internal/api/openapi.go`.
- `POST /api/v1/llm/chat`'s `tools` field expects each tool's schema under the snake_case key `input_schema`, because that endpoint's wire contract is now owned entirely by the vendored `github.com/pgEdge/pgedge-go-llm-lib` module (`llm.Tool.InputSchema` is tagged `json:"input_schema"`, mirroring Anthropic's own wire format).

The client takes the tool list verbatim from `/mcp/tools` and forwards it into the chat request. Because the field name doesn't match, `input_schema` silently unmarshals to empty for every tool (Go's `encoding/json` does not fuzzy-match `inputSchema` against a tag of `input_schema` — an underscore is not a case-only difference), and Anthropic rejects the resulting empty schema.

This was introduced by `ba54cc29` ("refactor(server): delete bespoke chat clients + dead embedding logger; slim chat to prompt helpers"), which migrated the chat/proxy plumbing onto the shared `pgedge-go-llm-lib` and its wire contract, without reconciling the pre-existing `inputSchema` convention the rest of the Workbench's REST API already used.

## Reproduction

1. `GET /api/v1/mcp/tools`, take any single tool object as-is.
2. `POST /api/v1/llm/chat` with `{"messages": [...], "tools": [<that tool object>]}`.
3. Observe the 400 error above, regardless of which tool is used.

## Impact

Live regression on `main`. Breaks Ask Ellie (and any other feature calling `/api/v1/llm/chat` with tools) for any interaction that requires a tool call — which is most real usage.

moizpgedge:
## Summary

- `POST /api/v1/llm/chat` and `/chat/stream` rejected every request that included a tool list with `anthropic (400): tools.0.custom.input_schema: Input does not match the expected shape`, breaking Ask Ellie and the Server, Query, and Alert analysis panels for any interaction requiring a tool call. This is 100% reproducible with any well-formed tool — it's systemic, not about a specific tool's schema.
- Root cause: `GET /api/v1/mcp/tools` correctly returns each tool's schema under the camelCase key `inputSchema`, matching the real MCP protocol's `mcp.Tool` struct — the same endpoint also serves genuine external MCP clients (e.g. Claude Desktop), so that field must stay camelCase there. The client took the tool list essentially verbatim and forwarded it as `tools` in the chat request, but that endpoint decodes its body via the vendored `github.com/pgEdge/pgedge-go-llm-lib` v0.1.1 package's `llm.Tool` struct, tagged `json:"input_schema"` (snake_case). Since `inputSchema` != `input_schema` for Go's `encoding/json` (an underscore is not a case-only difference, so there's no fallback match), every tool's `InputSchema` silently unmarshalled to `nil`, and Anthropic rejected the resulting empty schema.
- This is a **client-only** fix; nothing server-side changes, and `pkg/mcp/types.go` is untouched (renaming it would break the real MCP protocol surface for external clients).
- Adds a `normaliseTools` helper in `client/src/types/llm.ts`, mirroring the existing `normaliseMessages` helper added in the same historical wire-contract migration (the same migration that introduced this regression — the author normalised `messages` but missed `tools`). Applied at **both** independent places that build a `tools` list for `/api/v1/llm/chat`: `client/src/hooks/chat/chatAgenticLoop.ts` (the main Ask Ellie chat loop) and `client/src/utils/agenticLoop.ts` (the separate loop backing the Server/Query/Alert analysis panels) — both were broken, and both needed the same fix.

Closes #370

## Test plan

- [x] New unit tests for `normaliseTools` (`client/src/types/__tests__/llm.test.ts`): renames `inputSchema` → `input_schema`, preserves `name`/`description`, empty array, does not mutate input, preserves arbitrary schema shapes.
- [x] Regression-guard tests added at **both** call sites (`chatAgenticLoop.test.ts`, `agenticLoop.test.ts`) that assert on the actual outgoing request body's `tools` shape — no existing test did this (every mock fetch returned a scripted response regardless of the request body, and every Go-side test constructed already-correctly-shaped structs directly, bypassing JSON), which is why this shipped undetected.
- [x] Adversarially verified: reverted the `normaliseTools` call at both sites, confirmed the new regression-guard tests fail with the exact expected symptom, restored the fix, confirmed clean pass.
- [x] Full client suite passes (3484/3484), `npm run lint` clean (0 errors), `npm run build` clean.
- [x] 100% line coverage on all three touched source files.
- [x] Live-verified against the real Anthropic API (not mocked) through the actual running app: multi-turn tool-call round trips succeed end-to-end on **both** independent call sites (main chat invoking `list_connections`, and the analysis-loop path invoking `list_probes`), each confirmed both via direct API calls mirroring the real client request and by manually testing in the browser.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed chat requests that included tool definitions being rejected due to an invalid schema format.
  * Tool schemas are now sent using the expected `input_schema` format, improving reliability for Ask Ellie and analysis panels.

* **Documentation**
  * Added an unreleased changelog entry describing the chat request fix.

### DIFF
--- a/client/src/hooks/chat/__tests__/chatAgenticLoop.test.ts
+++ b/client/src/hooks/chat/__tests__/chatAgenticLoop.test.ts
@@ -760,7 +760,35 @@ describe('chatAgenticLoop', () => {
                 ]);
                 expect(body.system_prompt).toBe('Be helpful');
                 expect(body.system).toBeUndefined();
-                expect(body.tools).toEqual(params.availableTools);
+                expect(body.tools).toEqual(
+                    params.availableTools.map(t => ({
+                        name: t.name,
+                        description: t.description,
+                        input_schema: t.inputSchema,
+                    })),
+                );
+            });
+
+            it('sends tools with a snake_case input_schema key, not the internal camelCase inputSchema (issue #370)', async () => {
+                // Regression guard: the library `llm/proxy` chat endpoint
+                // decodes tools into a struct tagged `json:"input_schema"`.
+                // Sending the app's internal camelCase `inputSchema` field
+                // verbatim causes the schema to silently unmarshal to nil,
+                // and Anthropic rejects the resulting empty schema. See
+                // issue #370.
+                const mockFetch = createMockFetch([createTextResponse('Hi')]);
+                const params = createLoopParams({ fetchFn: mockFetch });
+
+                await runAgenticLoop(params);
+
+                const call = mockFetch.mock.calls[0];
+                const body = JSON.parse(call[1]?.body as string);
+
+                expect(body.tools.length).toBeGreaterThan(0);
+                for (const tool of body.tools) {
+                    expect(tool.input_schema).toBeDefined();
+                    expect(tool.inputSchema).toBeUndefined();
+                }
             });
 
             it('sends correct request body to tool call endpoint', async () => {
--- a/client/src/hooks/chat/chatAgenticLoop.ts
+++ b/client/src/hooks/chat/chatAgenticLoop.ts
@@ -24,7 +24,7 @@ import type {
     ToolCallResponse,
     ToolResult,
 } from '../../types/llm';
-import { normaliseMessages } from '../../types/llm';
+import { normaliseMessages, normaliseTools } from '../../types/llm';
 import type { APIMessage, ToolDefinition } from './chatTypes';
 
 /**
@@ -191,7 +191,7 @@ export async function runAgenticLoop(
             headers: { 'Content-Type': 'application/json' },
             body: JSON.stringify({
                 messages: normaliseMessages(currentMessages),
-                tools: availableTools,
+                tools: normaliseTools(availableTools),
                 system_prompt: systemPrompt,
             }),
             signal: abortSignal,
--- a/client/src/types/__tests__/llm.test.ts
+++ b/client/src/types/__tests__/llm.test.ts
@@ -1,0 +1,135 @@
+/*-------------------------------------------------------------------------
+ *
+ * pgEdge AI DBA Workbench
+ *
+ * Copyright (c) 2025 - 2026, pgEdge, Inc.
+ * This software is released under The PostgreSQL License
+ *
+ *-------------------------------------------------------------------------
+ */
+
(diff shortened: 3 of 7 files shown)


dpage (approved):
Reviewed the fix in full and it looks good: correct diagnosis, minimal scope, and the regression guards assert on the actual outgoing request body rather than a scripted mock, which closes the exact gap that let this ship undetected. I confirmed independently that these are the only two client paths that forward a `tools` list to `/api/v1/llm/chat`; the other direct callers (`useChartAnalysis`, `useQueryOverview`) send only `messages` and `system_prompt`, so they are correctly unaffected.

One non-blocking suggestion for a follow-up rather than this PR: `LLMTool.input_schema` and the `normaliseTools` parameter's `inputSchema` are both typed `unknown` (`client/src/types/llm.ts`). That is defensible since the field is arbitrary JSON Schema and the helper only shuttles it across, but the codebase already has a `ToolInputSchema` type, and using it here would give a little more compile-time safety at the call sites. Take it or leave it.

I have also opened a follow-up issue for the broader fragility this bug exposed (four hand-built `/api/v1/llm/chat` request bodies, each independently responsible for normalising `messages` and `tools`), since that is what allowed the `tools` field to be missed during the wire-contract migration in the first place.

Approving and merging.

Issue #27, opened 2026-03-11
SSL Settings Not Preserved After Adding Server Under Cluster


imtiazqa:
When adding a server under a cluster and explicitly setting the SSL option to Disabled, the value is not preserved. Upon reopening the server settings, the SSL settings field appears empty instead of showing "Disabled."

**Steps to Reproduce:**

- Navigate to the Cluster section.
- Add a new server under the cluster.
- In the server configuration, set SSL Settings = Disabled.
- Save the server configuration.
- Click on the Settings (gear) icon for the newly added server.
- Check the value of SSL Settings.

<img width="1598" height="864" alt="Image" src="https://github.com/user-attachments/assets/f7dfe590-c145-474a-8ab8-fe34a9addc2f" />

**Expected Result:**
The SSL Settings value should remain selected Disabled, reflecting the configuration selected during server creation.

Actual Result:
The SSL Settings field appears empty instead of showing Disabled.

imtiazqa:
This issue is fixed now . So marking it to done.

PR #378, opened 2026-07-29
fix(server): use Gemini embeddings in knowledgebase search

### LINKED ISSUE #337: Gemini provider: (1) knowledge base search silently falls back to OpenAI due to missing gemini_embedding column in search_knowledgebase.go; (2) session startup fails with 400 "empty Part" error
**Environment:**

Provider: gemini
Model: gemini-2.5-flash
Embedding model: gemini-embedding-001
KB file: kb-gemini-gemini-embedding-001.db (866MB)
Server version: v1.0.0 (unpatched in latest main as well)

**Error:**

ERROR: LLM chat request failed: Gemini API error (400):
GenerateContentRequest.contents[1].parts[0].data:
required oneof field 'data' must have one initialized field

**Steps to Reproduce:**

1. Configure server with provider: gemini, model: gemini-2.5-flash
2. Configure knowledgebase with gemini-embedding-001 and valid API key
3. Start the server — AI Overview generates a summary automatically on startup
4. Open chat (Ask Ellie)
5. Send any message — e.g. Look up in the knowledge base: what is connection pooling?
6. Chat fails immediately — blank/error response in UI, 400 error in server log

**Root Cause:**

AI Overview runs on startup and makes a Gemini API call to generate a summary. Gemini returns a response that includes an empty text Part alongside the function call. This gets stored in conversation history.

When the user sends a chat message, that stored history is replayed in the request. The replayed response contains an empty geminiPart{Text: ""}. Because geminiPart.Text is tagged json:"text,omitempty", an empty string serializes to {} — an empty object with no data field. Gemini rejects this:

type geminiPart struct {
    Text             string                  `json:"text,omitempty"`
    FunctionCall     *geminiFunctionCall     `json:"functionCall,omitempty"`
    FunctionResponse *geminiFunctionResponse `json:"functionResponse,omitempty"`
}
Affected File: server/src/internal/chat/llm_gemini.go

Fix:

// Fix 1 — line ~173, string case
case string:
    if content == "" {
        continue   // skip empty string messages
    }
    role := msg.Role
    ...

// Fix 2 — line ~211, TextContent case
case TextContent:
    if v.Text != "" {
        modelParts = append(modelParts, geminiPart{Text: v.Text})
    }

// Fix 3 — line ~243, map[string]any "text" case
if text, ok := itemMap["text"].(string); ok && text != "" {
    modelParts = append(modelParts, geminiPart{Text: text})
}

dpage:
## Summary

This addresses part (1) of #337: knowledgebase search never read the Gemini
embedding column, so a server configured with `embedding_provider: gemini`
silently searched against another provider's vectors.

`searchKB` selected only `openai_embedding`, `voyage_embedding` and
`ollama_embedding`, and its provider switch had no `gemini` case, so `gemini`
fell through to the `openai` default. A fallback chain then quietly
substituted whichever other provider's blob happened to be non-empty. Since
`cosineSimilarity` returns 0 whenever the two vectors differ in length, a
3072-dimension Gemini query compared against 1536-dimension OpenAI vectors
scored every row at zero, so results came back ranked arbitrarily rather than
by relevance; on a Gemini-only knowledgebase, where the OpenAI column is NULL,
every row was skipped and the search returned nothing at all.

The changes are:

- Knowledgebase files vary by vintage, and older ones have no
  `gemini_embedding` column, so the column list is now built dynamically from
  the columns the file actually carries; both old and new layouts work.
- Provider `gemini` now selects `gemini_embedding`.
- The silent cross-provider fallback is gone. When the configured provider's
  column is absent, or is empty for every matching chunk, the search returns
  an explicit error naming the provider and the knowledgebase file and
  explaining that the two must match, rather than degrading to irrelevant
  results.

Every column name reaching the generated SQL comes from a hard-coded
allow-list, never from configuration or user input; the project name and
version `IN` clauses are untouched and still bind their values as parameters.

## Scope: this does not close #337

Issue #337 covers two separate defects, and only the first is fixed here.

Part (2), the Gemini 400 `required oneof field 'data' must have one
initialised field`, is no longer in this repository at all. The bespoke chat
clients were deleted in ba54cc29 and the Gemini provider now lives in the
shared `pgedge-go-llm-lib`, which is where the remaining fragility sits. It is
fixed there by pgEdge/pgedge-go-llm-lib#27.

Note that the originally reported chain is already broken in v0.1.1, because
the library filters empty text parts when parsing responses, so the empty part
no longer enters conversation history. The request path is still unguarded
though, and a Gemini reply consisting only of an empty text part now parses to
an assistant message with zero content blocks, which the library turned into
an `{Text: ""}` placeholder and so reproduced the same 400 by a different
route.

Picking that up here needs a `pgedge-go-llm-lib` release carrying #27, and a
pin bump in `server`, `alerter` and `pkg`. That release is deliberately being
batched with other triage, so the bump is a known follow-up rather than part
of this PR. #337 should stay open until it lands.

## Test plan

- [x] New coverage for a knowledgebase carrying both Gemini and OpenAI
      columns, whose two columns rank the rows in opposite orders, proving the
      Gemini vectors are genuinely the ones used and that OpenAI still ranks
      correctly on the same file.
- [x] New coverage for a Gemini-only knowledgebase schema.
- [x] New coverage for a legacy schema with no `gemini_embedding` column
      searched as `gemini`, asserting the new explanatory error rather than a
      silent fallback or a bare SQLite "no such column".
- [x] New coverage for a knowledgebase with no embedding columns, an absent
      chunks table, and a file that is not a SQLite database.
- [x] Two existing tests asserted the fallback behaviour being removed; their
      assertions were rewritten rather than deleted, and a malformed-blob row
      was added to keep the deserialisation-skip path covered. The existing
      openai, voyage and ollama paths are otherwise unchanged.
- [x] `gofmt -l` clean, `go build ./...` clean, `make lint` reports 0 issues.
- [x] Per-function coverage on the new and modified code: `kbProviderColumn`
      and `kbProviderName` 100%, `kbPresentEmbeddingColumns` 93.3%, `searchKB`
      94.5%, all above the 90% floor. The uncovered statements are three
      unreachable defensive error branches.

`internal/tools` has two failing tests,
`TestStoreMemoryGeneratesEmbeddingIntegration` and
`TestRecallMemoriesGeneratesQueryEmbeddingIntegration`, both failing with
`expected 3 dimensions, not 4000`. I verified these fail identically on an
unmodified `main`, and they persist after recreating the local test database,
so they are pre-existing and unrelated to this change; they look like an
embedding output-dimensionality mismatch worth a separate issue.

Refs #337

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved knowledge-base search to use embeddings only from the configured provider.
  * Added clearer errors when the configured embedding column is missing, when all embeddings are empty, or when embeddings are malformed.
  * Removed unintended fallback behavior across providers and improved handling of older/incomplete and corrupted knowledge-base databases.

* **Tests**
  * Updated knowledge-base fixtures and added helpers for different embedding column layouts.
  * Expanded coverage for Gemini provider behavior, plus missing columns/tables and corrupt database scenarios.
  * Enhanced cases to ensure rows with unusable embeddings are skipped.

### DIFF
--- a/server/src/internal/tools/search_knowledgebase.go
+++ b/server/src/internal/tools/search_knowledgebase.go
@@ -303,6 +303,73 @@ func generateKBQueryEmbedding(ctx context.Context, serverCfg *config.Config, que
 	return vector32, embCfg.Provider, nil
 }
 
+// kbEmbeddingColumns is the fixed allow-list of embedding columns that
+// searchKB understands, in a stable order. Knowledgebase files vary by
+// vintage: older ones carry no gemini_embedding column at all, and a
+// provider-specific build may carry only one of these columns. Every
+// column name that reaches the generated SQL comes from this literal
+// list, never from user input or configuration, so interpolating the
+// selected names into the statement remains injection-safe.
+var kbEmbeddingColumns = []string{
+	"openai_embedding",
+	"voyage_embedding",
+	"ollama_embedding",
+	"gemini_embedding",
+}
+
+// kbProviderColumn maps a configured embedding provider onto the column
+// holding that provider's vectors. Unrecognized providers fall through
+// to openai, matching the historic behavior of this tool.
+func kbProviderColumn(provider string) string {
+	switch strings.ToLower(provider) {
+	case "voyage":
+		return "voyage_embedding"
+	case "ollama":
+		return "ollama_embedding"
+	case "gemini":
+		return "gemini_embedding"
+	default: // openai
+		return "openai_embedding"
+	}
+}
+
+// kbProviderName renders an embedding column as its provider name, for
+// use in user-facing error messages.
+func kbProviderName(column string) string {
+	return strings.TrimSuffix(column, "_embedding")
+}
+
+// kbPresentEmbeddingColumns probes the chunks table and returns the
+// subset of kbEmbeddingColumns that the knowledgebase file actually
+// carries, preserving the allow-list order. A missing or unreadable
+// chunks table surfaces as an error from the probe itself.
+func kbPresentEmbeddingColumns(db *sql.DB) ([]string, error) {
+	// LIMIT 0 fetches the column metadata without reading any rows.
+	rows, err := db.Query(`SELECT * FROM chunks LIMIT 0`)
+	if err != nil {
+		return nil, fmt.Errorf("failed to inspect chunks table: %w", err)
+	}
+	defer rows.Close()
+
+	names, err := rows.Columns()
+	if err != nil {
+		return nil, fmt.Errorf("failed to inspect chunks table: %w", err)
+	}
+
+	existing := make(map[string]bool, len(names))
+	for _, name := range names {
+		existing[name] = true
+	}
+
+	var present []string
+	for _, column := range kbEmbeddingColumns {
+		if existing[column] {
+			present = append(present, column)
+		}
+	}
+	return present, nil
+}
+
 func searchKB(kbPath string, queryEmbedding []float32, projectNames, projectVersions []string, topN int, provider string) ([]KBSearchResult, error) {
 	// Open database
 	db, err := sql.Open("sqlite", kbPath)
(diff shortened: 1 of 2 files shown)


AntTheLimey (approved):
No blocking issues; the provider-column resolution is right and the
fix is covered where it matters.

Three optional notes. There's no `docs/changelog.md` entry, and this
changes what `search_knowledgebase` returns once Gemini is configured:
previously a Gemini-configured server scored every chunk against
whichever other provider's column existed and returned a plausible but
meaningless ranking, and now it errors explicitly. A line under
Unreleased referencing #337 would give anyone tracking that issue
something to find.

`kbRankChunks` still drops `rows.Scan` failures with an uncounted
`continue`, which CodeRabbit flagged twice on earlier commits. If a
malformed metadata column fails every row's scan, `rowCount` stays 0,
the "all rows lack this provider's embedding" error never fires, and
the caller just sees no results: the same silent failure this PR
removes for provider mismatches. Counting scan failures alongside
`emptyProviderRows` would close it.

Separately, dimension safety is enforced per provider column but not
per model, so changing `embedding_model` under an unchanged
`embedding_provider: gemini` without rebuilding the knowledgebase
still gives all-zero similarities. That one is pre-existing and not
yours to fix here, but it is the same failure mode.

Everything else checks out. `gofmt` and `golangci-lint` are clean on
the touched package, the 23 KB test functions pass under `-race`
against real databases rather than mocks, every function you touched
measures at or above 90% line coverage, and server total coverage is
unchanged at 56.1%. No sibling PR touches these files and the branch
merges clean against current main.

PR #382, opened 2026-07-29
fix(server): widen memory embedding test fixture to halfvec(4000)

dpage:
## The problem

The `chat_memories` fixture used by the embedding integration tests in
`server/src/internal/tools/memory_tools_embedding_db_test.go` still declared its
embedding column as `vector(3)`, which was correct back when the mocked Gemini
endpoint's three-element response went into the database verbatim, but has been
wrong since the collector's migration v6 widened the embedding columns to
`halfvec(4000)` for multi-provider support. `memory.Store` now pads every
embedding out to `embeddingpkg.MaxDimensions`, which is 4000, on the insert path
in `server/src/internal/memory/store.go` and again on the query path, so
Postgres rejected the padded value outright with `ERROR: expected 3 dimensions,
not 4000 (SQLSTATE 22000)`. The sibling fixture in the memory package was
updated at the time and reads `halfvec(4000)` already; this one was missed.

The upshot is that both `TestStoreMemoryGeneratesEmbeddingIntegration` and
`TestRecallMemoriesGeneratesQueryEmbeddingIntegration` failed on any host where
pgvector is actually installed, which includes the development host.

## Why halfvec rather than a wider vector

**Corrected after review:** an earlier version of this description claimed
pgvector caps the `vector` type at 2000 dimensions. It does not. `vector` goes
to 16,000 dimensions, and 2,000 is the HNSW and IVFFlat index limit for
`vector`. Verified on pgvector 0.8.2: a `vector(4000)` column accepts the
store's padded value, and only an HNSW index on it is refused.

So `vector(4000)` would work for this fixture, which has no index. The reason
to use `halfvec` is simply that it matches production: `store.go` casts to
`::halfvec`, the canonical table carries a `halfvec_cosine_ops` index, and the
sibling memory-package fixture already reads `halfvec(4000)`. Matching the real
schema is the point, not any type limit.

## Why CI never caught this

These two tests have never actually executed in CI. The workflow's Postgres
service is a plain `postgres:<version>` image with no pgvector available, so
`CREATE EXTENSION IF NOT EXISTS vector` fails, the helper calls
`t.Skipf("pgvector extension unavailable")`, and the run reports green whilst
quietly skipping the whole seam. A follow-up PR adds pgvector to the CI Postgres
service so that this path is genuinely gated; no workflow files are touched
here, keeping the two concerns separate.

## Scope

This is a fixture and comment correction only: no production code changes, no
test restructuring, and no new tests. The stale comments asserting that the
mock's three-element vector matches the column type have been reworded to
explain the padding, since the mock returning three elements remains correct and
should stay.

## Verification

Run on the development host, where pgvector is installed, against the local
loopback Postgres:

- `gofmt -l` on the touched file is clean, and `go build ./...` succeeds.
- Both tests now PASS rather than skip:

  ```
  === RUN   TestStoreMemoryGeneratesEmbeddingIntegration
  --- PASS: TestStoreMemoryGeneratesEmbeddingIntegration (0.04s)
  === RUN   TestRecallMemoriesGeneratesQueryEmbeddingIntegration
  --- PASS: TestRecallMemoriesGeneratesQueryEmbeddingIntegration (0.02s)
  PASS
  ok      github.com/pgedge/ai-workbench/server/internal/tools     0.331s
  ```

- The whole `internal/tools` package is green, and coverage rises from 49.3% to
  51.1% now that these two tests execute rather than skip.
- `make lint` in `server` reports 0 issues.

https://claude.ai/code/session_015nPEZyXgN1zK7EuEXhfZvW

## Summary by CodeRabbit

* **Tests**
  * Updated memory embedding integration coverage to use production-compatible 4,000-dimensional half-precision vectors.
  * Clarified documentation describing how smaller embedding responses are padded to the required dimensionality before database operations.
  * Improved alignment between test scenarios and production embedding storage behavior.

### DIFF
--- a/server/src/internal/tools/memory_tools_embedding_db_test.go
+++ b/server/src/internal/tools/memory_tools_embedding_db_test.go
@@ -27,11 +27,13 @@ import (
 )
 
 // memoryEmbeddingTableDDL mirrors the chat_memories table that the memory
-// store reads from and writes to. The vector dimension is 3 so the mocked
-// Gemini endpoint can return a three-element vector and the store can
-// persist it without dimension mismatch. The table is created inside a
-// per-test private schema (see newMemoryEmbeddingStore) so it never
-// clobbers the canonical public.chat_memories shared with other tests.
+// store reads from and writes to. The column is halfvec(4000) because the
+// store pads every embedding to embedding.MaxDimensions before insert and
+// before querying, so a shorter vector column would reject the padded
+// value; halfvec rather than vector is required because pgvector caps the
+// vector type at 2000 dimensions. The table is created inside a per-test
+// private schema (see newMemoryEmbeddingStore) so it never clobbers the
+// canonical public.chat_memories shared with other tests.
 const memoryEmbeddingTableDDL = `
 CREATE TABLE chat_memories (
     id BIGSERIAL PRIMARY KEY,
@@ -40,7 +42,7 @@ CREATE TABLE chat_memories (
     category TEXT NOT NULL,
     content TEXT NOT NULL,
     pinned BOOLEAN NOT NULL DEFAULT FALSE,
-    embedding vector(3),
+    embedding halfvec(4000),
     model_name TEXT NOT NULL DEFAULT '',
     created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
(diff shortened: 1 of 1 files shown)


AntTheLimey (dismissed):
One correction before merge. The new `memoryEmbeddingTableDDL` comment
says "pgvector caps the vector type at 2000 dimensions". It does not:
`vector` goes to 16,000 dimensions, and 2,000 is the HNSW/IVFFlat index
limit for `vector`. On pgvector 0.8.6 a `vector(4000)` column accepts
the store's padded `::halfvec` value; only an HNSW index on it is
refused, and this fixture has no index. The real reason for halfvec is
the one `pkg/embedding/pad.go` and the memory-package fixture already
give: it matches the production column, the `::halfvec` casts in
`store.go`, and the `halfvec_cosine_ops` index. Suggest rewording that
clause and dropping the 2000 claim from the PR description too.

Otherwise good: both tests fail on `main` with `expected 3 dimensions,
not 4000` and pass here against pgvector 0.8.6.

Merge order: #383 carries this same commit under a different SHA. Land
this first and rebase #383 with the reword, or #383 will restore the
current wording.

dpage:
You're right, and thanks for checking it properly rather than taking the claim at face value. `vector` goes to 16,000 dimensions and 2,000 is the HNSW/IVFFlat index limit; I verified on pgvector 0.8.2 here that a `vector(4000)` column accepts the store's padded value and only an HNSW index on it is refused.

Reworded the fixture comment to give the actual reason (it matches production: `store.go` casts to `::halfvec` and the canonical table carries a `halfvec_cosine_ops` index) and dropped the 2000 claim from the PR description as well.

PR #383, opened 2026-07-29
ci: run Postgres services on pgvector images so vector-gated tests actually execute

dpage:
## Why

Several of our integration tests create the pgvector extension and skip
themselves when it turns out to be missing, with messages such as
`pgvector extension unavailable: %v`. Every CI Postgres service, however,
used a plain `postgres:<version>` image, which carries no pgvector, so
**every one of those tests has silently skipped in CI since it was
written**; a skip reads as a pass in the job summary line, and nobody
looks past a green tick. That is precisely how the stale
`embedding vector(3)` fixture in the server memory tests survived on
`main`, against a store that pads every vector to 4000 dimensions, until
it was found by running the suite on a dev host that does have pgvector
installed.

## What changed

The four Postgres service containers now use the corresponding
`pgvector/pgvector:pg<version>` tags:

- `.github/workflows/ci-server.yml`
- `.github/workflows/ci-collector.yml`
- `.github/workflows/ci-alerter.yml`
- `.github/workflows/ci-e2e.yml` (fixed `pg16`)

Nothing else about those service blocks changed, and no Postgres or Go
version was bumped. The pgvector images are built directly on the
official images (their Dockerfile is `FROM postgres:$PG_MAJOR-$DEBIAN_CODENAME`
plus a build of the extension), so the entrypoint, the
`POSTGRES_PASSWORD`/`POSTGRES_DB` environment, `pg_isready` health
checks and port mappings all carry over unchanged. The tags `pg14`
through `pg18` all exist, covering every matrix value, and no explicit
`CREATE EXTENSION` step is required because each test creates the
extension itself against the container's superuser role.

## Tests that start executing in CI

I ran all five pgvector-gated files for real against a local Postgres 18
with pgvector 0.8.2, which is something CI has never done. Every one of
these now executes rather than skipping, and all of them pass:

`server/src/internal/memory/store_db_test.go`

- `TestStore_Store`, `TestStore_GetByID`, `TestStore_Delete`,
  `TestStore_DeleteByID`, `TestStore_UpdatePinned`,
  `TestStore_GetPinned`, `TestStore_ListByUser`, `TestStore_Search`

`server/src/internal/tools/memory_tools_embedding_db_test.go`

- `TestStoreMemoryGeneratesEmbeddingIntegration`,
  `TestRecallMemoriesGeneratesQueryEmbeddingIntegration`

`collector/src/database/migration_v6_test.go`

- `TestMigrationV6_FreshSchemaUsesHalfvec`,
  `TestMigrationV6_UpgradesLegacyVectorColumns`,
  `TestMigrationV6_Idempotent`,
  `TestUpgradeEmbeddingColumn_AbsentAndUnexpectedType`,
  `TestUpgradeEmbeddingColumn_WrongHalfvecWidth`,
  `TestUpgradeEmbeddingColumn_StepFailures`

`collector/src/database/schema_idempotency_test.go`

- The pgvector-conditional assertions inside `TestMigrateFreshDatabase`,
  `TestMigrateTwiceIsIdempotent`, `TestMigratePartialState`,
  `TestRunPgVectorSetup_SuccessAndError` and
  `TestRunChatMemoryEmbeddingSetup_SuccessAndError`, including the
  `anomaly_embeddings` foreign-key check that previously skipped with
  "pgvector not available; FK absent by design"

`alerter/src/internal/database/queries_full_integration_test.go` and the
neighbouring `anomaly_queries_full_integration_test.go`

- `TestStoreAnomalyEmbeddingAndFindSimilar`,
  `TestFindSimilarAnomaliesScanError`, plus the `FindSimilarAnomalies`
  paths gated behind `pgvectorAvailable`

## Tests that still skip

None of the pgvector-gated tests skip once the extension is present.
The remaining skip conditions in these files are `SKIP_DB_TESTS` and an
unset `TEST_AI_WORKBENCH_SERVER`, and all three Go workflows already
export `TEST_AI_WORKBENCH_SERVER` for the coverage step, so the gap is
genuinely closed rather than merely narrowed. A verbose run of all four
affected packages locally produced no `--- SKIP` lines at all.

The one secondary gate worth naming is `pgvectorHasHalfvec` in the
collector migration tests, which skips when the installed pgvector
predates 0.7.0. The floating `pgXX` tags currently resolve to pgvector
0.8.5, so halfvec is present on every matrix entry and the gate passes;
if a future pgvector image ever regressed that, those tests would skip
again, and pinning the tags to an explicit `0.8.x-pgXX` would be the
remedy.

## Fixture drift

None beyond what #382 already fixes. Everything passed once the
extension was available, so this PR uncovers no further drift and
introduces no production changes; the diff is four YAML lines.

## Stacking

This PR is **stacked on #382** and must be merged after it, or rebased
onto `main` once #382 lands. Without the halfvec fixture fix in #382, the
server memory tests newly enabled here would fail immediately, which is
rather the point.

## Verification boundary

Locally I ran the four affected Go packages against a real pgvector
instance and validated each changed workflow with `yaml.safe_load`, but
CI cannot be proven from a dev host: the local runs only cover Postgres
18, and the Postgres 14 to 17 matrix entries, along with the E2E job's
now-pgvector-capable schema path, are exercised for the very first time
by this PR's own CI run. That run is the real proof, not my local
output.

https://claude.ai/code/session_015nPEZyXgN1zK7EuEXhfZvW

## Summary by CodeRabbit

- **Tests**
  - CI and end-to-end test environments now use a pinned PostgreSQL image with vector support.
  - Integration tests can enable the vector extension when available, improving coverage for vector-enabled database configurations.
  - Local end-to-end runs now use the same vector-enabled PostgreSQL setup as CI.

- **Documentation**
  - Updated testing guidance to cover additional workflows, supported Go versions, and PostgreSQL provisioning details.

### DIFF
--- a/.claude/golang-expert/testing-strategy.md
+++ b/.claude/golang-expert/testing-strategy.md
@@ -742,22 +742,32 @@ func executeCommand(cmd string) { ... }
 - `.github/workflows/ci-collector.yml`
 - `.github/workflows/ci-server.yml`
 - `.github/workflows/ci-alerter.yml`
+- `.github/workflows/ci-e2e.yml`
 - `.github/workflows/ci-client.yml`
 - `.github/workflows/ci-docs.yml`
 
-All workflows:
+The Go workflows:
 
-1. **Matrix**: Go 1.23, 1.24
-2. **Services**: PostgreSQL 14, 15, 16, 17, 18
-3. **Steps**: Build, test with coverage, lint, upload coverage HTML
-4. **Artifacts**: Coverage reports retained 7 days
+1. **Matrix**: Go 1.26.2
+2. **Steps**: Build, test with coverage, lint, upload coverage HTML
+3. **Artifacts**: Coverage reports retained 7 days
+
+Only four of them declare a PostgreSQL service:
+`ci-collector.yml`, `ci-server.yml` and `ci-alerter.yml` across
+PostgreSQL 14, 15, 16, 17 and 18, and `ci-e2e.yml` on 16 alone.
+`ci-client.yml` and `ci-docs.yml` need no database.
 
 ### CI Example (Server)
 
 ```yaml
 services:
   postgres:
-    image: postgres:${{ matrix.postgres-version }}
+    # pgvector, not plain postgres: the server, collector and alerter
+    # all have tests gated on the vector extension, and on a plain
+    # image those tests skip silently rather than fail. Pin the
+    # pgvector version so a new upstream release cannot change what CI
+    # tests underneath us.
+    image: pgvector/pgvector:0.8.6-pg${{ matrix.postgres-version }}
     env:
       POSTGRES_PASSWORD: postgres
       POSTGRES_DB: postgres
--- a/.github/workflows/ci-alerter.yml
+++ b/.github/workflows/ci-alerter.yml
@@ -24,7 +24,7 @@ jobs:
 
         services:
             postgres:
-                image: postgres:${{ matrix.postgres-version }}
+                image: pgvector/pgvector:pg${{ matrix.postgres-version }}
                 env:
                     POSTGRES_PASSWORD: postgres
                     POSTGRES_DB: postgres
--- a/.github/workflows/ci-alerter.yml
+++ b/.github/workflows/ci-alerter.yml
@@ -24,7 +24,7 @@ jobs:
 
         services:
             postgres:
-                image: pgvector/pgvector:pg${{ matrix.postgres-version }}
+                image: pgvector/pgvector:0.8.6-pg${{ matrix.postgres-version }}
                 env:
                     POSTGRES_PASSWORD: postgres
                     POSTGRES_DB: postgres
--- a/.github/workflows/ci-collector.yml
+++ b/.github/workflows/ci-collector.yml
@@ -24,7 +24,7 @@ jobs:
 
         services:
             postgres:
-                image: postgres:${{ matrix.postgres-version }}
+                image: pgvector/pgvector:pg${{ matrix.postgres-version }}
                 env:
                     POSTGRES_PASSWORD: postgres
                     POSTGRES_DB: postgres
--- a/.github/workflows/ci-collector.yml
+++ b/.github/workflows/ci-collector.yml
@@ -24,7 +24,7 @@ jobs:
 
         services:
             postgres:
-                image: pgvector/pgvector:pg${{ matrix.postgres-version }}
+                image: pgvector/pgvector:0.8.6-pg${{ matrix.postgres-version }}
                 env:
                     POSTGRES_PASSWORD: postgres
                     POSTGRES_DB: postgres
--- a/.github/workflows/ci-e2e.yml
+++ b/.github/workflows/ci-e2e.yml
@@ -31,7 +31,7 @@ jobs:
 
         services:
             postgres:
-                image: postgres:16
+                image: pgvector/pgvector:pg16
                 env:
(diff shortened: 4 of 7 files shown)


AntTheLimey (dismissed):
Two things to fix before merge; the workflow change itself is right and
CI proves the server and collector vector tests now execute.

**Blocking**

1. The alerter tests this PR lists as now executing still skip in its
   own CI. `pgvectorAvailable` at
   `alerter/src/internal/database/queries_full_integration_test.go:389`
   checks `pg_extension` (installed), but the package's only
   `CREATE EXTENSION IF NOT EXISTS vector` is at line 410, after the
   gate. On a fresh pgvector container the extension is available but
   not installed, so the gate is false. The Alerter PG 18 and PG 14
   jobs for a97d5824 both log `pgvector not available; skipping
   embedding tests` and `--- SKIP` for
   `TestStoreAnomalyEmbeddingAndFindSimilar` and
   `TestFindSimilarAnomaliesScanError`. I reproduced it locally; they
   pass only where someone has already run `CREATE EXTENSION`, which is
   what a dev-host database has. Either move the `CREATE EXTENSION`
   ahead of the gate here, or drop the alerter claims from the body and
   commit message and open a follow-up.

2. `.claude/golang-expert/testing-strategy.md:760` reproduces the
   service block this PR changes, still as `image: postgres:...`.
   CLAUDE.md requires the KB to move in the same change; please switch
   that line so the next workflow copied from the KB does not
   reintroduce the skip.

**Optional**

- `e2e/run-local.sh:61` still runs `postgres:16`, so the local E2E
  harness and `ci-e2e.yml` now take different schema paths. Same
  one-token change.
- The body says the tags resolve to pgvector 0.8.5; `pg18` shipped
  0.8.6 when I checked. That drift is the case for pinning
  `0.8.x-pgNN`, given the workflows already pin actions by SHA. Your
  call.
- While in `testing-strategy.md`: line 750 says Go 1.23/1.24 (workflows
  use 1.26.2), and lines 741-746 list `ci-client.yml` and `ci-docs.yml`
  as having Postgres services (they do not) and omit `ci-e2e.yml`.
  Pre-existing, cheap to fix in the same edit.

Merge order: base is `main` and #382 is still open, so this carries
#382's fixture as a different SHA. Merge after #382, or rebase once it
lands, carrying across any comment reword from #382.

dpage:
All three fixed, and both optional items taken.

**Blocking 1.** You're right, and I reproduced the mechanism directly: on a freshly created database `pg_extension` reports `f` for `vector` whilst `pg_available_extensions` reports `t`, so the gate returned false before anything could install it, and the tests only passed where someone had already run `CREATE EXTENSION` by hand, which is exactly the state of a dev-host database. `pgvectorAvailable` now performs the install it was implicitly relying on, which is both the real question and what `createAnomalyEmbeddingsTable` does anyway.

**Blocking 2.** `testing-strategy.md` updated. Whilst in there I also took your third optional point: Go 1.23/1.24 corrected to 1.26.2, `ci-client.yml` and `ci-docs.yml` no longer listed as carrying PostgreSQL services, and `ci-e2e.yml` added with its single PG 16 service.

**Optional, both done.** `e2e/run-local.sh` moves to the same image so the local harness and `ci-e2e.yml` stop diverging. And I've taken the pinning suggestion: `0.8.6-pg{14..18}` all exist, so every service is now pinned to 0.8.6 rather than tracking the floating `pgNN` tag. You were right about the drift; `pg18` is on 0.8.6 today.

Rebased onto current `main`, which now carries #410, #412 and #415. #382 is still open, so the merge-order point stands: this should land after it.

PR #385, opened 2026-07-29
Guard against non-explainable statements in the Query Plan panel

### LINKED ISSUE #368: Query Plan panel shows raw syntax error for non-explainable statements (VACUUM, ANALYZE, etc.)
## Symptom

Clicking into a Top Queries row whose captured statement is a bare
utility command (e.g. `VACUUM`, from a scheduled `VACUUM (ANALYZE,
VERBOSE)` job) opens the Query Detail view's Query Plan panel showing
a raw, unfriendly error:

```
Text plan: Error: Query error: syntax error at or near "VACUUM";
JSON plan: Error: Query error: syntax error at or near "VACUUM"
```

## Root cause

`useQueryPlan`'s `fetchExplain` (`client/src/hooks/useQueryPlan.ts:112-159`)
builds its request by literally concatenating `EXPLAIN` with the
captured query text, e.g.:

```ts
explainQuery = format === 'json'
    ? `EXPLAIN (VERBOSE, FORMAT JSON) ${query}`
    : `EXPLAIN ${query}`;
```

`pg_stat_statements` tracks utility/DDL statements as well as
SELECT/DML, so a row can legitimately contain bare text like `VACUUM`,
standalone `ANALYZE`, `REINDEX ...`, `CLUSTER ...`, etc. PostgreSQL's
`EXPLAIN` only supports SELECT, INSERT, UPDATE, DELETE, MERGE, VALUES,
EXECUTE, DECLARE, CREATE TABLE AS, and CREATE MATERIALIZED VIEW (plus
WITH-prefixed forms of these) — never utility commands. Wrapping one
of those in `EXPLAIN` is invalid SQL, so PostgreSQL correctly rejects
it with a syntax error, which the panel surfaces verbatim instead of
recognizing the statement type up front.

This isn't specific to any one cron job — it will happen for any
monitored database with routine maintenance activity (autovacuum,
manual VACUUM/ANALYZE/REINDEX) showing up in Top Queries.

## Proposed fix

Detect non-explainable statement types before attempting EXPLAIN, and
show a clear message (e.g. "Query plans aren't available for
maintenance statements like VACUUM, ANALYZE, or REINDEX") instead of
firing the request and surfacing a raw Postgres syntax error.

dpage:
## Summary

- `useQueryPlan` built its request by concatenating `EXPLAIN` with the
  captured statement text, so a Top Queries row holding a utility
  statement (a bare `VACUUM` from a scheduled maintenance job, a
  standalone `ANALYZE`, `REINDEX`, and so on) produced invalid SQL and
  the panel surfaced PostgreSQL's raw `syntax error at or near "VACUUM"`
  verbatim.
- The hook now decides up front whether `EXPLAIN` can accept the
  statement and, when it cannot, sets an explanatory message without
  issuing the request at all. The check is an allowlist of the leading
  keywords `EXPLAIN` supports (`SELECT`, `INSERT`, `UPDATE`, `DELETE`,
  `MERGE`, `VALUES`, `EXECUTE`, `DECLARE`, `WITH`, `CREATE TABLE ... AS`
  and `CREATE MATERIALIZED VIEW`) rather than a denylist of utility
  commands, so it fails safe: anything unrecognised yields the friendly
  message instead of a raw server error, and there is no catalogue of
  utility commands to maintain as PostgreSQL grows new ones.
- Leading whitespace, `--` and `/* */` comments, and opening parentheses
  are stripped before the first keyword is inspected, since captured
  `pg_stat_statements` text is often prefixed with a framework tag
  comment. `CREATE TABLE ... AS SELECT` stays explainable whilst a plain
  `CREATE TABLE` with a `GENERATED ALWAYS AS` column does not.

No component changes were needed: `QueryPlanPanel` already renders an
error with no accompanying plan as an `info` alert, so the message reads
as "not available" rather than as a failure, sitting alongside the
existing friendly handling of parameterised queries.

## Test plan

- [x] `npx vitest run src/hooks/__tests__/useQueryPlan.test.ts` — 79
  passed (11 pre-existing plus 68 new). The new cases cover the helper
  directly across case, whitespace, comment, parenthesis, CTAS and
  materialised-view forms plus the utility statements from the issue,
  and cover the hook asserting that a non-explainable statement produces
  the message and that the API mock is never called.
- [x] `npx vitest run --coverage` on the modified hook — 97.91% lines
  (94/96), 97.82% branches; the two uncovered lines are pre-existing and
  every new line is covered, comfortably over the project's 90% floor.
- [x] Full client suite: 171 files, 3556 tests passed.
- [x] `npm run lint` — 0 errors, 40 warnings, all pre-existing
  non-null-assertion warnings across the test tree (confirmed by blame).
- [x] `npm run build` succeeds.

Closes #368

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Query Plan behavior for unsupported PostgreSQL statements by showing a clear unavailable message instead of a database error.
  * Avoids unnecessary plan requests for unsupported or planless statements and clears stale plan output.
  * Improved handling of valid and invalid JSON query-plan responses.

* **Tests**
  * Expanded coverage for statement classification, short-circuiting, and varied EXPLAIN response formats.

* **Documentation**
  * Updated the Unreleased changelog with the Query Detail Query Plan panel fix.


AntTheLimey (changes requested):
One fix before merge: `stripLeadingNoise` skips a block comment by
finding the first `*/`, but PostgreSQL nests `/* */` comments, so a
nested leading comment closes a level early and the classifier reads
the wrong token. `classifyExplainSupport('/* outer /* inner */ SELECT
1 */ VACUUM')` returns `plan`, the untouched text goes to the server,
and the panel shows the raw `syntax error at or near "VACUUM"` this
PR removes. The mirror case `'/* a /* b */ VACUUM */ SELECT 1'`
returns `unsupported` for a plain SELECT. Tracking nesting depth in
the `/*` branch fixes both; a test for each direction would pin it.

Optional: `CREATE_TABLE_AS_PATTERN` requires the query keyword
straight after `AS`, so `CREATE TABLE t AS /* c */ SELECT 1` is
classified `unsupported` though EXPLAIN accepts it. Harmless
direction, so a follow-up if you want it.

Everything else checks out: every supported EXPLAIN form I ran through
`classifyExplainSupport` classifies correctly, the CTAS versus
`GENERATED ALWAYS AS` distinction holds, a non-explainable statement
never issues a request and clears stale plan state (confirmed in a
browser against a stack built from the branch: VACUUM and ANALYZE
rows show the banner and `pg_stat_statements` records no EXPLAIN for
them), coverage on
`useQueryPlan.ts` measures 99.11% lines, lint is clean, and the
changelog entry matches. No other open PR touches these files and the
branch is level with main.

dpage:
Pushed 9b7feb79 addressing the review.

- `stripLeadingNoise` now skips block comments with a new `skipBlockComment` that tracks `/* */` nesting depth, so `/* outer /* inner */ SELECT 1 */ VACUUM` classifies as unsupported and `/* a /* b */ VACUUM */ SELECT 1` as plan. An unterminated nested comment is treated as comment to the end of the text, so it can neither hang nor leak a token. Tests cover both directions, three-deep nesting, `/*/* a */*/` and the unterminated cases.
- The optional CTAS point is done too: `CREATE_TABLE_AS_PATTERN` is replaced by `isCreateTableAs`, which checks the `CREATE ... TABLE` prefix and then, for each `AS` token, runs the same noise stripping on the remainder before testing for `SELECT`, `VALUES`, `TABLE`, `EXECUTE` or `WITH`. `CREATE TABLE t AS /* c */ SELECT 1` and the `-- c` line-comment form now classify as plan, whilst `GENERATED ALWAYS AS (...)`, `GENERATED ALWAYS AS /* c */ (1) STORED` and identity columns stay unsupported. Every `AS` is scanned rather than the first because column definitions may legitimately contain one before the query.

Gates: lint clean; `useQueryPlan.ts` at 99.25% lines and 100% branches (the one uncovered line predates this branch); full client suite 3653 tests passing with coverage totals above the thresholds.

AntTheLimey (approved):
Confirmed fixed. `skipBlockComment` tracks nesting depth correctly —
both repro cases from this morning now classify right
(`/* outer /* inner */ SELECT 1 */ VACUUM` to unsupported, `/* a /* b
*/ VACUUM */ SELECT 1` to plan), and I ran it against nested-string,
four-deep-nesting, interleaved `--`/`/* */`, and stray-`*/` cases on
top of your 32 new tests without finding a gap. `useQueryPlan.ts`
coverage is 99.25% lines with one pre-existing uncovered line, full
`make coverage` passes, lint is clean, and typecheck is flat against
main (570/570). The CTAS-comment optional point is fixed too. Nothing
else outstanding; approving.

PR #395, opened 2026-08-07
Remove dead code identified by reachability analysis

dpage:
## What this is

Following on from the Codacy review, I looked at what the tree contains
that nothing calls. Running `deadcode` from each service entry point
found **94 functions unreachable** from the collector, server and
alerter binaries.

Cross-checking against the **v1.0.0** tag settled whether any of it was
work-in-progress: every one of the 94 was already unreachable at that
release two months ago, so none is recent code awaiting a caller. Worth
noting the converse, though: four functions that *were* dead at v1.0.0
have since been adopted (`EstimateTokens`, `EstimateTotalTokens`,
`GetBriefDescription`, `HasToolResults`), which is a decent argument for
removing on evidence rather than case-by-case optimism.

This PR takes the mechanically safe subset. Two follow-ups are noted at
the bottom.

## Scope

| | Lines |
|---|---|
| Production Go removed | 698 (47 functions) |
| Test Go removed | 2,110 |
| Client | 117 |
| **Total** | **2,925 deletions, 22 insertions** (the insertions are the changelog) |

Two removals went beyond single functions:

- The alerter's `secretManager` had no callers, and with it gone the
  `SecretManager` interface in `interfaces.go` had no implementation and
  no users either, so both are deleted.
- The resource `Registry` type, its constructor and its methods were
  used only by their own tests. `Resource` and `Handler`, declared in
  the same file, **stay**: the MCP resource definitions in
  `pg_system_info.go` use them.

On the client, the unused `HeaderStatusIndicator` component goes, along
with `@dnd-kit/sortable` and `@dnd-kit/utilities`. Only `@dnd-kit/core`
is imported and it remains.

## How the deletions were validated

Each function was deleted and verified **individually**, and kept only
where the resulting breakage was confined to test functions whose own
names target the deleted symbol. Anything else was reverted
automatically.

That guard is the most important thing in this PR. A first pass without
it deleted 262 test functions and 11,382 lines, including entire
integration suites such as the auth handler login, rate-limiting and
secure-cookie tests, and the alert `Unacknowledge` tests. Those cover
live code and broke only because they call a dead method during cleanup
or use a private setter as an injection seam. I had allowed a test to be
removed when its name mentioned the *receiver type*, which is far too
loose; the rule now requires the function name itself.

## Coverage

Essentially unmoved, because the deleted code and its deleted tests
cancel out:

| Module | `main` | This branch |
|---|---|---|
| collector | 85.5% | 85.5% |
| server | 55.5% | 55.3% |
| alerter | 65.9% | 65.6% |

No new code is added, so there is nothing new to cover and the 90%
floor for new and modified code does not bite. The server sitting below
the 60% repository goal is pre-existing rather than introduced here,
and closing that 4.7-point gap is a substantial separate piece of work
rather than something to bolt onto a deletion.

## Verification

- `gofmt`, `golangci-lint` and `go vet` clean on all three modules.
  `golangci-lint` caught one genuine follow-on, an orphaned `reWord`
  regexp left behind by deleting `CountWords`, now removed.
- Full suites pass on collector and alerter. The server's only failures
  are the two pre-existing `vector(3)` fixture failures from #337,
  which reproduce identically on unmodified `main`.
- Client: lint clean, production build succeeds, 3,488 tests across 171
  files pass.

One note for anyone re-running this locally: the suites must be run with
`-p=1`, as the sub-project Makefiles do. I first ran them without it and
got a spray of `deadlock detected` and `relation already exists`
failures, because the integration tests all build the same schema in one
shared database. Incidentally `collector/Makefile` does **not** set
`-p=1` where `server/Makefile` does, which looks like an oversight worth
a separate look.

## Deliberately left for follow-up

- **42 further dead functions** whose tests need editing rather than
  deleting. These cluster usefully: the ten notification-channel CRUD
  methods in the alerter all share one
  `TestNotificationQueriesReturnErrorOnClosedPool`, the `internal/logging`
  façade shares `TestLogLevelFiltering`, and several are test seams like
  `AuthHandler.Close` and `AlertHandler.setAlertResolver`.
- **The two superseded handler constructors**, `NewConnectionHandler`
  and `NewNotificationChannelHandler`. `cmd/mcp-server` calls the
  `...WithSecurity` variants; the test suites still use the old ones, so
  32 tests need repointing at the live constructors first.
- **Three test helpers in non-test files** (`NewTestClient`,
  `NewTestDatastoreWithSecret`, `NewTestDatastore`). They compile into
  the shipped binaries but are used across package boundaries by tests,
  so they cannot simply become `_test.go` files; the right fix is
  extracting them into a test-only package, which is import churn rather
  than deletion.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed looping `metric_staleness` alerts: failed or unevaluable metrics no longer clear alerts, with improved availability checks, cooldown protection, and per-probe alert handling.
- **Removed**
  - Removed numerous unused internal APIs, notification secret-management utilities, configuration-saving helpers, resource-management features, and dedicated tests.
  - Removed obsolete status-panel UI, drag-and-drop dependencies, logging, validation, caching, analytics, formatting, and legacy access helpers.
- **Documentation**
  - Updated the unreleased changelog with alerting fixes and removed functionality.


AntTheLimey (dismissed):
Request changes, for one small fix.

Blocking: `alerter/internal/cron` coverage drops from 95.2% to 78.9%.
Deleting `TestValidate` and `TestDefaultParser` along with the
package-level `Parse` and `Validate` wrappers also removed the only
test of the package-level `Matches`, which `engine/blackouts.go` still
calls via `cronMatches`; it is now at 0%. The same two tests were the
only callers of `(*Parser).Validate`, which now has no caller anywhere,
so it should have gone with the wrapper. Suggested fix: delete
`(*Parser).Validate`, and either add one test for the free `Matches` or
have `blackouts.go` call `cron.DefaultParser.Matches` and drop that
wrapper as well.

Optional: the NOTE comment in `alerter/src/internal/database/queries.go`
still lists `GetLatestMetricValue` among the functions in
`metric_queries.go`, which this PR deletes.

Everything else checked out. Every removed symbol was confirmed dead
repo-wide, including string-keyed registry lookups. The removed auth,
validation and secrets code all had live equivalents on the real paths
(`ParseQueryTimeSilent`, `ValidateStringsInSet`, per-tool type
assertions, `crypto.DecryptPassword` in the notifications manager),
and the SQL identifier guards are untouched. The client removals are
clean: nothing imports `HeaderStatusIndicator` or the two `@dnd-kit`
packages, and `npm ci`, build, tests and lint pass with coverage
unchanged. Build, vet, gofmt and golangci-lint are clean in all three
Go sub-projects; no knowledge-base file or OpenAPI entry references the
removed code; the changelog matches the diff.

Merge order: none of the nine other open PRs touching these files
calls a removed symbol, so this merges first without breaking anything.
#390, #412 and #397 will need a textual rebase on adjacent lines. One
note shared with #397: this PR removes `Reset` and `GetSummary` from
the compactor's `Analytics` and #397 removes its other two readers,
none of which has a production caller today, so after both land
`compactor.go` records into a struct nothing can read. The #397 review
asks for that to be resolved there.

dpage:
Fixed, and your diagnosis was exactly right: 78.9% total, `(*Parser).Validate` at 0% with no caller anywhere, and the package-level `Matches` at 0% despite `blackouts.go` reaching it through `cronMatches`.

Took both halves of your suggestion rather than either/or, because they apply to different functions. `(*Parser).Validate` is deleted, which is the same reachability argument the rest of the PR rests on. The package-level `Matches` stays, since it is genuinely live, and now has direct tests instead: weekday ranges, timezone conversion, the empty-timezone default, parse failures, and one case worth calling out.

An unknown timezone does not error; `Matches` falls back to UTC silently. So a blackout schedule with a mistyped timezone quietly evaluates in UTC rather than failing, which seemed worth pinning with a test and a comment whilst I was in there. I had assumed it errored and the test told me otherwise.

The package is back to 100%, and I've bumped the changelog's function count from 47 to 48.

PR #396, opened 2026-08-07
Repoint handler tests at the WithSecurity constructors

dpage:
The follow-up to #395, covering the one piece I deliberately left out of
the mechanical pass.

## Why this needed its own PR

`NewConnectionHandler` and `NewNotificationChannelHandler` are
unreachable from the server binary: `cmd/mcp-server/handlers.go` wires
both handlers through the `...WithSecurity` variants. They survived #395
because **49 test call sites** still used the shorter forms, and deleting
the constructors there would have taken out test suites covering live
handler code, which is exactly the failure mode #395 was tightened to
avoid.

## Why the rewrite is safe

The two constructor pairs are identical apart from a single line:

- short form: `hostValidator: DefaultHostValidator()`
- `WithSecurity`: `hostValidator: NewHostValidator(allowInternal, allowedHosts, blockedHosts)`

and `DefaultHostValidator()` is defined as *exactly*:

```go
func DefaultHostValidator() *HostValidator {
    return NewHostValidator(false, nil, nil)
}
```

So rewriting `NewConnectionHandler(a, b, c)` to
`NewConnectionHandlerWithSecurity(a, b, c, false, nil, nil)` is
precisely equivalent. No test changes behaviour, and nothing needed
reinterpreting: every field the short constructor set is set identically
by the longer one. That is what makes this mechanical rather than a
judgement call, which is the thing I wasn't willing to assume without
checking.

With the short forms gone, `DefaultHostValidator` had no production
caller either, so it goes too. Its two incidental uses in
`host_validation_test.go` (in `TestHostValidator_ValidatePort` and
`TestHostValidator_InternalNetworksList`, which use it only as a
convenient default) now call `NewHostValidator(false, nil, nil)`
directly. `TestDefaultHostValidator` is removed along with the function
it existed to test.

## Scope

| | Lines |
|---|---|
| Production Go removed | 31 (3 functions) |
| Tests | −67 / +51 (net −16) |
| Changelog | +12 |

49 call sites rewritten across 7 test files. No test was deleted except
`TestDefaultHostValidator`.

## Verification

- `gofmt` clean, `golangci-lint` reports **0 issues**, `go vet` clean.
- Full server suite passes with `-race -p=1`. The only failures are the
  two pre-existing `vector(3)` fixture failures from #337, which
  reproduce identically on unmodified `main`.
- **Server coverage 55.5%, matching `main` exactly** (#395 leaves it at
  55.3%). Nothing to top up; no new code is added, so the 90% floor for
  new and modified code does not apply.
- `deadcode` confirms no cascade: the count from the server binary drops
  from 65 to 62, exactly the three functions removed here, with nothing
  newly stranded in the touched area.

## Relationship to #395

Cut from `main`, not stacked on `remove-dead-code`, because a branch
based on a non-`main` branch triggers no CI workflows at all. I verified
the two changesets touch **entirely disjoint files** (52 files in #395,
11 here, zero overlap), so they can merge in either order without
conflict.

## Still outstanding after this

The **42 remaining dead functions** whose tests need editing rather than
deleting, and the **three test helpers** (`NewTestClient`,
`NewTestDatastoreWithSecret`, `NewTestDatastore`) that compile into the
shipped binaries but are used across package boundaries by tests, so
they want extracting into a test-only package rather than deleting.

## Summary by CodeRabbit

- **Removed**
  - Removed legacy connection and notification-channel constructors, along with the default host-validation helper.
  - Handler creation now uses security-aware configuration, while host validation requires explicit settings.

- **Documentation**
  - Added an Unreleased changelog entry documenting the removed APIs and migration path.

- **Tests**
  - Updated validation, routing, authentication, and RBAC coverage to use the current security-aware configuration without changing expected behavior.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -391,6 +391,18 @@ project adheres to
   call sites all go together. `EnableAnalytics` defaulted to false, so
   no deployment was collecting these figures in any case.
 
+### Removed
+
+- Remove the superseded `NewConnectionHandler` and
+  `NewNotificationChannelHandler` constructors, along with the
+  `DefaultHostValidator` helper that only they used. The server has
+  wired both handlers through the `NewConnectionHandlerWithSecurity` and
+  `NewNotificationChannelHandlerWithSecurity` variants for some time, so
+  the shorter forms were unreachable in production whilst 49 test call
+  sites still used them. Those call sites now use the `WithSecurity`
+  constructors directly, passing the same host-validation settings the
+  removed helper supplied, so test behaviour is unchanged.
+
 ### Security
 
 - Ignore a blank password when updating a database connection, so an
--- a/server/src/internal/api/connection_handlers.go
+++ b/server/src/internal/api/connection_handlers.go
@@ -36,16 +36,6 @@ type ConnectionHandler struct {
 	visibilityListerFn func() auth.ConnectionVisibilityLister
 }
 
-// NewConnectionHandler creates a new connection handler
-func NewConnectionHandler(datastore *database.Datastore, authStore *auth.AuthStore, rbacChecker *auth.RBACChecker) *ConnectionHandler {
-	return &ConnectionHandler{
-		datastore:     datastore,
-		authStore:     authStore,
-		hostValidator: DefaultHostValidator(),
-		rbacChecker:   rbacChecker,
-	}
-}
-
 // NewConnectionHandlerWithSecurity creates a new connection handler with custom security settings
 func NewConnectionHandlerWithSecurity(datastore *database.Datastore, authStore *auth.AuthStore,
 	rbacChecker *auth.RBACChecker, allowInternal bool, allowedHosts, blockedHosts []string) *ConnectionHandler {
--- a/server/src/internal/api/connection_handlers_test.go
+++ b/server/src/internal/api/connection_handlers_test.go
@@ -26,7 +26,7 @@ import (
 )
 
 func TestNewConnectionHandler(t *testing.T) {
-	handler := NewConnectionHandler(nil, nil, nil)
+	handler := NewConnectionHandlerWithSecurity(nil, nil, nil, false, nil, nil)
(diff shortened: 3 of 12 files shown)


AntTheLimey (approved):
No blocking issues; this does what it says, and I confirmed the
substitution is exact rather than taking it on trust: all 49 rewritten
call sites pass `(false, nil, nil)`, which is what `DefaultHostValidator`
was defined as, so no test reaches a host-validation or RBAC branch it
did not reach before.

The one thing to act on sits in #420, not here. The two merge clean at
the git level and then do not compile, in either order: #420 adds
`query_execute_rollback_test.go`, which calls the now-deleted
`NewConnectionHandler` three times, and this PR never touches that file,
so git gives no warning and CI cannot catch it either, since Codacy is
the only required check on main and the second PR to merge is never
forced to rebase. Rather than make it a merge-order rule someone has to
remember, I have asked #420 to switch those three calls to
`NewConnectionHandlerWithSecurity(ds, store, checker, false, nil, nil)`
now: that compiles against today's main and against main after this
lands, so the ordering stops mattering. Nothing for you to change here.
#411 conflicts only textually in `docs/changelog.md`, adjacent Unreleased
entries. #386 merges clean.

Optional: `TestNewConnectionHandler` now builds through
`NewConnectionHandlerWithSecurity` but keeps its old name and a
`t.Fatal` message reading "NewConnectionHandler returned nil", next to
an existing `TestNewConnectionHandlerWithSecurity`. CodeRabbit raised
the same nitpick.

Everything else checks out: `gofmt`, `golangci-lint`, `go vet` and
`go build` all clean with no stranded imports, no caller of the three
deleted constructors survives anywhere across the four sub-projects,
no knowledge-base file names them, no endpoint changed so OpenAPI is
correctly untouched, and server coverage is 56.1% on both main and the
branch with `internal/api` moving 53.1% to 53.2% as dead code leaves
the denominator.

PR #398, opened 2026-08-07
Stop the AuthHandler login rate limiter on shutdown

dpage:
Found while triaging the dead-code findings, and flagged in #397.

## The defect

`NewAuthHandler` creates a rate limiter for itself, covering total login
attempts per IP:

```go
totalRateLimiter: auth.NewRateLimiter(1, 20), // 20 total login requests per minute per IP
```

`auth.NewRateLimiter` starts a background cleanup goroutine (`go
rl.cleanupLoop()`). Only `AuthHandler.Close` stops it, and **nothing
called `Close`**: `SetupHandlers` constructs the handler inside the
closure it returns and discards the reference, so neither the server nor
anything else could reach it.

The result is a goroutine plus its map of recorded attempts living for
the remaining life of the process.

## Honest severity

**Bounded, not growing.** `SetupHandlers(deps)` is called once from
`server.go:453` and its closure runs once, so it is one goroutine and one
small map, and the process exits at shutdown anyway. This is not an
unbounded leak and I don't want to oversell it.

It is still worth fixing, for three reasons:

1. `Server.Close` already stops the overview generator, the shared rate
   limiter, the auth store and the datastore. This is an inconsistency in
   that pattern rather than a deliberate exemption.
2. `Close` exists and its doc comment documents exactly this ownership
   split, explaining that the caller-supplied `rateLimiter` is
   deliberately *not* stopped there. Not calling it is an oversight.
3. Because `Close` looks unreachable, static analysis reports it as dead
   code. That is how it nearly got **deleted** rather than wired up: it
   was in the deferred set in #395 and I came close to removing it in
   #397 before checking what it did.

## The fix

`HandlerDependencies` gains an optional `RegisterCloser func(func())`.
`SetupHandlers` uses it to hand `authHandler.Close` back to the server,
and `Server.Close` runs the collected closers alongside the resources it
already stops.

Registration is mutex-guarded, because `SetupHandlers` runs on the HTTP
server's goroutine whilst `Close` may run from the signal handler.
`runHandlerClosers` clears the list, so a second shutdown pass does not
stop the same resource twice. The callback is optional, so existing
callers and tests that only exercise route registration are unaffected.

This also gives any future handler owning background work a place to hand
that ownership back, rather than each one inventing its own escape hatch.

## Verification

- Six new tests: closer registration including the nil guard, run-once
  and list-clearing behaviour, the no-closers case (which is what the CLI
  subcommands hit, since they never start the listener), concurrent
  registration under `-race`, and the actual defect, that `SetupHandlers`
  registers a closer for the auth handler.
- **Both new functions are at 100% statement coverage.**
- Server coverage rises from 55.5% to **55.9%**.
- `gofmt` clean, `golangci-lint` **0 issues**, `go vet` clean.
- Full server suite passes with `-race -p=1`; the only failures are the
  two pre-existing `vector(3)` fixture failures from #337, which
  reproduce identically on unmodified `main`.
- `deadcode` now reports `AuthHandler.Close` as **reachable**, dropping
  the server's unreachable count from 65 to 64. That is the useful
  confirmation the wiring actually took effect.

204 insertions, no deletions.

## Merge order

Cut from `main`. The only file shared with #395, #396 and #397 is
`docs/changelog.md`, and this entry goes under `### Fixed` whilst theirs
go under `### Removed`, so it should merge cleanly in any order. No code
files overlap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server shutdown to reliably stop authentication background processes and release associated resources.
  * Ensured cleanup tasks run safely once, including during concurrent shutdown activity and late registration.

* **Tests**
  * Added coverage for cleanup registration, execution, concurrency, repeated shutdown behavior, and optional or missing cleanup handlers.

* **Documentation**
  * Documented the authentication cleanup improvements in the changelog.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -142,6 +142,19 @@ project adheres to
   the `pg_stat_database` probe reports on `current_database()` only.
   (#403)
 
+- Stop the login rate limiter that `AuthHandler` creates for itself
+  when the server shuts down. `NewAuthHandler` starts a background
+  cleanup goroutine for an internal rate limiter covering total login
+  attempts per IP, and only its `Close` method stops that goroutine.
+  Nothing called `Close`, because `SetupHandlers` built the handler
+  inside a closure and discarded the reference, so the goroutine and
+  its map of recorded attempts survived for the remaining life of the
+  process. `HandlerDependencies` now carries a `RegisterCloser`
+  callback, which the server uses to collect cleanup functions from the
+  handlers it wires and to run them from `Close`, alongside the
+  overview generator, the shared rate limiter, the auth store, and the
+  datastore it already stopped.
+
 - Fix every chat request that included a tool list failing with
   `anthropic (400): tools.0.custom.input_schema: Input does not
   match the expected shape`, which broke Ask Ellie and the Server,
--- a/server/src/cmd/mcp-server/handler_closers_test.go
+++ b/server/src/cmd/mcp-server/handler_closers_test.go
@@ -1,0 +1,136 @@
+/*-------------------------------------------------------------------------
+ *
+ * pgEdge AI DBA Workbench
+ *
+ * Copyright (c) 2025 - 2026, pgEdge, Inc.
+ * This software is released under The PostgreSQL License
+ *
+ *-------------------------------------------------------------------------
+ */
+package main
+
+import (
+	"net/http"
+	"sync"
+	"testing"
+
+	"github.com/pgedge/ai-workbench/server/internal/config"
+)
+
+// TestRegisterHandlerCloser_RecordsClosers verifies that registered
+// closers accumulate in order and that a nil closer is ignored rather
+// than stored, since running a nil func would panic during shutdown.
+func TestRegisterHandlerCloser_RecordsClosers(t *testing.T) {
+	s := &Server{}
+
+	s.registerHandlerCloser(nil)
+	if len(s.handlerClosers) != 0 {
+		t.Fatalf("nil closer should be ignored, got %d entries",
+			len(s.handlerClosers))
+	}
+
+	s.registerHandlerCloser(func() {})
+	s.registerHandlerCloser(func() {})
+	if len(s.handlerClosers) != 2 {
+		t.Errorf("expected 2 registered closers, got %d",
+			len(s.handlerClosers))
+	}
+}
+
+// TestRunHandlerClosers_RunsEachOnce verifies that every registered
+// closer runs, and that the list is cleared afterwards so a second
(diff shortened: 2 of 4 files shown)


AntTheLimey (approved):
No blocking issues. I reproduced the leak directly: calling
`SetupHandlers` without a `RegisterCloser` leaves the `AuthHandler`'s
internal login-rate-limiter cleanup goroutine running indefinitely,
and wiring the closer returns the goroutine count to baseline. The
mechanism itself is idempotent and clean under `-race`.

One thing worth your attention, not blocking. On the actual SIGTERM
path this `Stop` may not run at all. `setupShutdownHandler` leaves the
`closer` slot of `shutdownDeps` nil on purpose, per its own comment
about past double-close panics, so `runShutdown` drains, skips the nil
closer and calls `exit(0)`, while main's `defer server.Close()` races
it independently once `RunHTTP` returns. I measured `ListenAndServe`
returning 72us into a 544ms drain, so with in-flight requests the
defer wins comfortably; on an idle shutdown, where `Shutdown` returns
almost at once, the two are within microseconds of each other and
`exit(0)` can win. The practical cost is nil, since the process is
leaving and the goroutine outlives it by microseconds either way. But
the tests establish that the closer mechanism works when something
calls it, not that production calls it, so the fix is unproven exactly
where it is aimed. Now that `runHandlerClosers` is guarded by
`closersDrained`, the double-close hazard that justified the nil slot
looks gone, so wiring the closer would make this deterministic.

Also optional: the new tests never assert the security property.
`TestSetupHandlers_RegistersAuthHandlerCloser` checks that the closer
exists and does not panic; nothing drives `handleLogin` past the
total-request limit after `Close` to confirm the 429 still fires, and
nothing asserts the separately-owned failed-login lockout limiter is
left alone. A test doing both would close that gap.

Everything else checks out. A security pass over the auth surface came
back clear: stopping the limiter cannot fail open, because
`RateLimiter.Stop` only closes the pruning goroutine's channel under a
`sync.Once` and never touches the mutex-guarded attempts map, and the
listener is closed before any cleanup can run, so no login request
arrives unprotected. `gofmt` and `golangci-lint` are clean, the nine
new tests pass under `-race`, `registerHandlerCloser` and
`runHandlerClosers` are both 100% covered, and package coverage rises
from 56.1% to 56.5%. `Close` itself reports 0% on this branch and on
main alike, an inherited gap rather than a regression. CodeRabbit's
post-drain registration race is addressed by content. No endpoint
changed, so OpenAPI is correctly untouched, and #396 merges clean
against this branch in either order.

Issue #409, opened 2026-08-11
Alerter reimplements the server's rate derivation, and its deltas are probe-interval dependent


dpage:
## Description

The alerter does not use the server's derived-metric facility. It
reimplements rate and delta logic in roughly 1200 lines of
hand-written SQL in
`alerter/src/internal/database/metric_registry.go`. Searching
`alerter/src` for the server's `_per_sec` machinery returns nothing.

The server's implementation lives in
`server/src/internal/metrics/query.go`. The two are free to drift, and
already have.

## Why this matters now

The most significant divergence: the alerter's deltas are
**per-sample**, not per-second. Every delta threshold is therefore
implicitly scaled by the probe's `collection_interval_seconds`. An
operator retuning a probe interval silently rescales the alert
threshold, with no warning and no change to the rule.

`checkpoint_warning` is the clearest example — its unit is labelled
`checkpoints`, but the value is checkpoints per probe interval.

This also means the correctness work planned for the server's rate
derivation (per-dimension deltas before aggregation, the `stats_reset`
guard, gap handling) will not benefit the alerter unless it is
duplicated a second time.

## Suggested direction

Either extract the rate derivation into a package both services
consume, or normalise the alerter's deltas to per-second so thresholds
become interval-independent. The former is more work but stops the
drift; the latter is a smaller change that fixes the immediate hazard.

At minimum, thresholds whose unit is per-probe-interval should say so
in the rule metadata, and changing a probe interval should warn about
affected rules.

## Related smaller items

`required_extension` is stored on every rule but never enforced by the
alerter; the only references are the scan targets at
`alerter/src/internal/database/queries.go:142` and `:420`. Rules gated
on `system_stats`, `spock` or `pg_stat_statements` are evaluated
unconditionally and rely on the metric query returning zero rows.
Benign today, but it makes "extension missing" indistinguishable from
"condition resolved", which interacts badly with the false-resolution
behaviour tracked separately.

`table_bloat_ratio` (`collector/src/database/schema.go:2700`) is
misnamed and duplicative. It computes `n_dead_tup / n_live_tup * 100`,
which is a dead-tuple ratio rather than bloat, and overlaps
`dead_tuple_ratio` (`:2724`) which uses `n_dead / (n_live + n_dead)`.
Two rules, near-identical quantities, different denominators,
thresholds of 50 and 20, and different categories.

dpage:
## Proposal

[...]

Issue #408, opened 2026-08-11
Anomaly detection ignores time-aware baselines, is disabled for 14 metrics, and mixes up databases


dpage:
## Description

Four defects in the anomaly detection path, all confirmed empirically
against a live PostgreSQL 16 instance. Together they mean seasonality
is never applied, detection silently never runs for a large minority
of metrics, and multi-database servers score one database against
another's baseline.

Tests are on branch `claude/dashboard-chart-stats-review-0iq76e` in
`alerter/src/internal/engine/audit_defects_test.go`.

## Hourly and daily baselines are computed and never read

`calculateBaselines` writes `period_type` rows for `hourly` (24 per
metric per connection) and `daily` (7) at `baselines.go:115,118`.
`GetMetricBaselines` (`anomaly_queries.go:231`) orders by
`period_type`, which is TEXT, so alphabetically `all` sorts before
`daily` and `hourly`. `anomalies.go:160` takes `baselines[0]`, so the
`all` row always wins.

`GetMetricBaselines` returns `[all, daily, hourly]` regardless of
insert order. End to end: with a warm, wide `all` baseline (sigma
1000) and a warm, tight `hourly` baseline (sigma 1) for the current
hour, a value 400 sigma from the hourly baseline produces zero
candidates. Equally, a cold `all` baseline suppresses detection even
when a mature hourly baseline exists.

Every time-aware baseline is written, indexed, retained, and ignored.
The tuned per-period warmup thresholds at `config.go:307-314` are
likewise unreachable. A metric with a strong diurnal cycle — session
count, query duration, temp files — has its overnight trough and
midday peak scored against a single 7 day grand mean, which inflates
the standard deviation and blinds detection to genuine within-cycle
deviation.

Fix is small: order by an explicit period preference, or select the
baseline matching the current hour and weekday.

Tests: `TestAuditC6BaselineOrderingPrefersAll`,
`TestAuditC6DetectAnomaliesIgnoresTimeAwareBaselines`, and the
failing-by-design `...Demo` variant.

## Detection is permanently disabled for 14 of 32 registry metrics

`GetHistoricalMetricValues` errors when `historicalSQL` is empty
(`metric_queries.go:204`). `calculateBaselines` then falls back to
`calculateGlobalBaselinesFallback` (`baselines.go:66,253`), which
builds a baseline from `GetLatestMetricValues` — typically one row per
connection, giving `sample_count = 1`, `stddev = 0`, and a NULL
[...]

Issue #407, opened 2026-08-11
Alerts flap, latch, or falsely resolve because missing data is treated as resolution


dpage:
## Description

Several rules fire or clear incorrectly. The common root cause is that
`checkAlertResolved` (`alerter/src/internal/engine/cleanup.go:48-53`
and `:74-78`) treats any query error or missing row as "condition
resolved", combined with metric queries whose windows are only
marginally wider than the probe interval.

Tests are on branch `claude/dashboard-chart-stats-review-0iq76e`.

## `cache_hit_ratio_low` returns every row in the window

`metric_registry.go:703` is the only registry entry with no
latest-sample reduction — no `DISTINCT ON`, no `rn = 1`, no `MAX`. It
returns every delta row in the 15 minute window, about three rows per
database at the 300 second `pg_stat_database` interval. Its siblings
`deadlocks_delta` and `temp_files_delta` both reduce with `GROUP BY`.

`evaluateRuleForAllConnections` (`thresholds.go:53`) fires if **any**
row violates, while `checkAlertResolved` (`cleanup.go:59`) breaks on
the **first** row. There is no `ORDER BY`, so rows arrive in window
function order, oldest first.

The consequence depends on which interval violates:

- Violation in the newest interval — the evaluator fires and the
  cleaner clears on byte-identical data. Demonstrated end to end over
  two cycles, producing 2 fire and 2 clear notifications.

- Violation in the oldest interval — the cleaner sees the violating
  row first, so the alert latches instead of flapping.

Fix: add a latest-sample reduction, matching the sibling metrics.

Tests: `TestAuditC8CacheHitRatioReturnsEveryDeltaRow`,
`TestAuditC8CacheHitRatioFiresAndClearsOnSameData`.

## `slow_query_count` latches on a lifetime average

`metric_registry.go:886` counts distinct queryids where
`pg_stat_statements.mean_exec_time` exceeds 1000 ms, with no
`total_exec_time` or `calls` delta. That column is a cumulative
average since `pg_stat_statements_reset()` and never decays.

Twelve queryids with a constant `calls` count — that is, never
executed during the window — and a lifetime mean of 4200 ms all get
counted, giving a value of 12 against the seeded threshold of 10. The
collector fetches up to 1000 statements, so the count is not bounded
below the threshold. It clears only on a manual stats reset or
eviction.

This is the alerter's one genuine instance of the cumulative-counter
problem found in the dashboard charts.

[...]

Issue #406, opened 2026-08-11
Five built-in alert rules can never fire


dpage:
## Description

Five seeded alert rules are incapable of firing under any conditions.
Each has a different root cause. All were confirmed empirically except
where noted.

Tests are on branch `claude/dashboard-chart-stats-review-0iq76e` in
`alerter/src/internal/database/audit_defects_test.go`.

## 1. `wal_archive_failed` queries a table that does not exist

Severity critical. Seeded at `collector/src/database/schema.go:2720`;
the metric `pg_stat_archiver.failed_count_delta`
(`alerter/src/internal/database/metric_registry.go:659`) selects `FROM
metrics.pg_stat_archiver`.

Applying the real collector `SchemaManager` to a fresh database yields
36 `metrics.*` tables, none named `pg_stat_archiver`, and no views.
Archiver columns live on `metrics.pg_stat_wal`
(`schema.go:858,876`). Every evaluation raises `relation
"metrics.pg_stat_archiver" does not exist (SQLSTATE 42P01)`, which is
swallowed at `thresholds.go:48`.

WAL archiving failure has never been detectable.

Fix: repoint the metric at `metrics.pg_stat_wal`.

Tests: `TestAuditC2ArchiverTableDoesNotExist`,
`TestAuditC2ArchiverRuleErrorIsSwallowed`.

## 2. `transaction_wraparound` evaluates a hardcoded constant

Severity critical. Seeded at `schema.go:2712` as `age_percent > 75`.
The metric (`metric_registry.go:1043`) joins `pg_stat_all_tables` and
`pg_settings`, discards both, and returns `50.0::float`.

With two tables seeded at 1 and 5e9 live tuples, the metric still
returns exactly 50.0.

`metrics.pg_stat_all_tables` has no frozen-xid column, so the true
value is not computable from the table the query was written against.
The data does exist in `metrics.pg_database.age_datfrozenxid`
(`schema.go:1250`, 300 second interval).

One caveat worth noting: a user-configured per-connection threshold
below 50 would flip this from never firing to firing permanently.

Test: `TestAuditC3TransactionWraparoundReturnsConstant`.

## 3. `high_max_connections` and `connection_utilization` expire after an hour

Seeded at `schema.go:2687-2688`. Both require a `metrics.pg_settings`
row newer than one hour (`metric_registry.go:46,80`). The
`pg_settings` probe is change-tracked
(`collector/src/probes/pg_settings_probe.go:80`) and skips the write
[...]

PR #410, opened 2026-08-11
Add regression tests documenting alerter defects

dpage:
## Summary

Adds test-only coverage that empirically verifies ten claimed defects
in the alerting subsystem. No production code is changed; the tests
establish a baseline so that each fix has something concrete to break.

This came out of a review of dashboard charts showing cumulative
rather than point-in-time statistics. The chart findings are tracked in
#400, #401, #402, #403 and #404. The alerter findings verified here are
tracked in #405, #406, #407, #408 and #409.

## Two test styles

`TestAudit*` tests assert the current, defective behaviour with a doc
comment stating what the behaviour should be. They pass in CI today
and fail once a fix lands, which is the signal to update them
alongside the fix.

`TestAudit*Demo` tests assert the correct behaviour and therefore fail
against current code. They are skipped unless `ALERTER_DEFECT_DEMO=1`
so CI stays green. Four exist, covering the staleness loop, baseline
ordering, the fallback baseline warmup gate, and the slow query
lifetime average.

Running the demo tests produces:

```
=== RUN   TestAuditC1StalenessAlertFireClearLoopDemo
    audit_defects_test.go:331: alert rows = 3, want 1 for a continuously stale probe
    audit_defects_test.go:334: active alerts = 0, want 1; the probe is still stale
    audit_defects_test.go:337: received an unexpected extra notification: type=alert_clear
--- FAIL: TestAuditC1StalenessAlertFireClearLoopDemo (0.08s)
=== RUN   TestAuditC6DetectAnomaliesIgnoresTimeAwareBaselinesDemo
    audit_defects_test.go:859: expected an anomaly candidate from the mature hourly baseline, got none
--- FAIL: TestAuditC6DetectAnomaliesIgnoresTimeAwareBaselinesDemo (0.04s)
=== RUN   TestAuditC7FallbackBaselineCanNeverWarmDemo
    audit_defects_test.go:1151: persisted baseline has a NULL earliest_sample_at, so the warmup gate can never admit it
--- FAIL: TestAuditC7FallbackBaselineCanNeverWarmDemo (0.04s)
=== RUN   TestAuditC9SlowQueryCountDemo
    audit_defects_test.go:919: slow_query_count = 1 for a query that did not run during the window, want 0
--- FAIL: TestAuditC9SlowQueryCountDemo (0.03s)
```

## What was verified

Confirmed: the `metric_staleness` fire and clear loop and its missing
cooldown guard (#405); the missing `metrics.pg_stat_archiver` table,
the hardcoded `transaction_wraparound` metric, the `pg_settings` one
hour window, and the NULL CPU column consequence on Linux (#406); the
`slow_query_count` lifetime average (#407); the anomaly baseline
ordering bug and the ignored `database_name` (#408).

Partially confirmed, with corrections to the original claims:

- The fallback baseline mechanism is real, but the count was wrong.
  The registry has 32 entries with 14 empty `historicalSQL`, not 12 of
  34; 34 is the number of seeded alert rules, not registry metrics.

- `cache_hit_ratio_low` has two failure modes rather than one. A
  violation in the newest interval flaps; a violation in the oldest
  interval latches, because the cleaner breaks on the first row of an
  unordered result.

Not provable here: that `system_stats` leaves `processor_time_percent`
NULL on Linux. The extension cannot be installed in this environment,
so the premise is confirmed from the upstream C source while the
consequence is covered by a table-driven test.

## Testing

Tests use the existing integration harness (`TEST_AI_WORKBENCH_SERVER`,
`NewTestDatastore`, `newEngineSpockTestEnv`, `newDetectAnomaliesEnv`)
against a live PostgreSQL 16 instance. The full alerter suite passes
under `-race -p=1`.

```
$ cd alerter/src && gofmt -l .    # no output
$ go vet ./...                    # no output
$ cd alerter && make coverage
ok   .../alerter/internal/database          coverage: 83.6%
ok   .../alerter/internal/engine     4.476s coverage: 57.0%
```

One caveat to flag: `make test-all` fails at `lint` for a pre-existing
environmental reason, reproduced on a clean tree with no changes:

```
Error: can't load config: the Go language version (go1.25) used to build
golangci-lint is lower than the targeted Go version (1.26.2)
```

`golangci-lint 2.5.0` is built with `go1.25.1` while the modules target
`go 1.26.1`. The linter cannot run in this environment at all; this is
unrelated to the change but means lint has not been verified.

The project's 90% coverage floor applies to new and modified
production code. This change adds no production code.

https://claude.ai/code/session_01L2VcS2NfYNkagXRo7khohq

---
_Generated by [Claude Code](https://claude.ai/code/session_01L2VcS2NfYNkagXRo7khohq)_

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive integration coverage for database metrics, audit data, anomaly baselines, cache-hit ratios, slow-query detection, and alert behavior.
  * Added scenarios covering stale data, missing database context, archiver errors, transaction age, CPU metrics, and notification handling.
  * Included opt-in demonstration tests for corrected slow-query and alerting behavior.
* **Documentation**
  * Documented current edge cases and expected behavior through representative test scenarios.

### DIFF
--- a/alerter/src/internal/database/audit_defects_test.go
+++ b/alerter/src/internal/database/audit_defects_test.go
@@ -1,0 +1,984 @@
+/*-------------------------------------------------------------------------
+ *
+ * pgEdge AI DBA Workbench
+ *
+ * Copyright (c) 2025 - 2026, pgEdge, Inc.
+ * This software is released under The PostgreSQL License
+ *
+ *-------------------------------------------------------------------------
+ */
+package database
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// This file holds verification tests for an audit of the alerter
+// subsystem. Each test documents an audit claim and pins the behavior
+// that the code actually exhibits today. Where the audit claim
+// describes a defect, the test asserts the CURRENT (defective)
+// behavior and the comment states what the behavior SHOULD be, so the
+// suite stays green until the defect is fixed; a fix will make the
+// pinned assertion fail, which is the intended signal to update the
+// test alongside the fix.
+//
+// Tests whose name ends in "Demo" assert the CORRECT behavior instead
+// and therefore fail against the current code. They are skipped unless
+// ALERTER_DEFECT_DEMO=1 is set, so CI stays green while the defect can
+// still be demonstrated on demand.
+
+// defectDemoEnabled skips the calling test unless ALERTER_DEFECT_DEMO
+// is set. Demo tests assert the behavior the audit says the code
+// should have, so they fail while the defect is present.
+func defectDemoEnabled(t *testing.T) {
+	t.Helper()
+	if os.Getenv("ALERTER_DEFECT_DEMO") == "" {
+		t.Skip("set ALERTER_DEFECT_DEMO=1 to run the defect demonstration")
+	}
+}
+
+// auditDefectsSchema mirrors the production collector schema for the
+// columns the metric registry queries actually read. The table set is
+// deliberately faithful on one point: metrics.pg_stat_wal exists (it
+// carries the archiver columns in production) and
+// metrics.pg_stat_archiver does NOT, because the collector never
+// creates such a table.
+const auditDefectsSchema = `
+DROP SCHEMA IF EXISTS metrics CASCADE;
+DROP TABLE IF EXISTS metric_baselines CASCADE;
+DROP TABLE IF EXISTS connections CASCADE;
+
+CREATE TABLE connections (
+    id SERIAL PRIMARY KEY,
+    name VARCHAR(255) NOT NULL,
+    enabled BOOLEAN NOT NULL DEFAULT TRUE,
+    is_monitored BOOLEAN NOT NULL DEFAULT TRUE
+);
+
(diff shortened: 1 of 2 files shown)


AntTheLimey (approved):
Approve.

Traced all ten defect claims to the production code they pin and
checked each doc comment's "should be" against the linked issue. Ran
the suite against Postgres 16 twice in each mode: every `TestAudit*`
passes and the four Demo tests skip; under `ALERTER_DEFECT_DEMO=1` the
same four fail with output matching the PR body's transcript. `gofmt`,
`go vet` and `golangci-lint` are clean, and every test tears down its
own schema.

Optional: `.claude/golang-expert/testing-strategy.md` does not yet
mention the `ALERTER_DEFECT_DEMO` gate or the paired `TestAudit*` and
`...Demo` convention. A short addition would make the pattern
discoverable for the next defect-regression PR.

Merge order: I applied these two test files on top of #412 and #418.
Against #412, `TestAuditC1StalenessAlertFireClearLoop` and
`TestAuditC1StalenessPathSkipsCooldownGuard` fail and the Demo variant
starts passing. Against #418, the C2, C3, C4 and C5 tests fail, with
C3 failing on a missing `metrics.pg_database` relation in the audit
fixture rather than a wrong value. Neither fix touches these files, so
whichever of the three lands last needs a follow-up commit updating
the named tests, and for #418 the fixture schema, or the alerter suite
on `main` goes red.

PR #411, opened 2026-08-11
fix(llm): honour configured max_tokens on AI analysis paths

### LINKED ISSUE #399: AI analysis paths hardcode a 512-token output cap, producing empty summaries with local reasoning models
## Summary

The estate summary, server-info AI analysis, and alerter reasoning paths
each hardcode a small output-token cap that ignores the configured
`llm.max_tokens`. With a modern reasoning model served locally (for
example, via llama.cpp), the cap is consumed entirely by the model's
thinking block and the summary comes back empty.

This was reported while building an all-in-one Docker stack for the
workbench against a local llama.cpp server; it reproduces on the
`v1.0.0` GA release and is still present on `main` at `19c645d`.

## Affected code

The following constants are compile-time values with no configuration
override:

| Path | Constant | Value |
|---|---|---|
| Estate summary | `server/src/internal/overview/generator.go:35` | `llmMaxTokens = 512` |
| Server-info analysis | `server/src/internal/api/server_info_handlers.go:34` | `llmAnalysisMaxTokens = 512` |
| Alerter reasoning | `alerter/src/internal/llm/reasoning.go:48` | `MaxTokens: pgllm.Int(500)` |

The configurable `llm.max_tokens` setting defaults to `4096`
(`server/src/internal/config/config.go:363,616`) and drives the main
agentic path, but none of the three paths above consult it. Setting
`llm.max_tokens` in the server configuration therefore has no effect on
the estate summary.

## Why the failure mode is an empty summary

Two behaviours compound to turn a tight cap into a total failure:

- On an OpenAI-compatible endpoint, `max_tokens` bounds the total
  generated tokens; a reasoning model's thinking tokens count against
  that budget. Models such as Qwen3, the DeepSeek-R1 distills, and
  gpt-oss routinely spend more than 512 tokens reasoning before emitting
  any answer text.

- `extractTextFromResponse` in `server/src/internal/overview/generator.go:557`
  concatenates only blocks whose type is `pgllm.BlockText`. When the
  budget is exhausted during reasoning, the response contains no text
  block, so the function returns an empty string rather than a truncated
  summary.

The caller treats that empty string as a successful generation, so the
failure surfaces as a blank panel with no error.

## Steps to reproduce

In the following steps, a local llama.cpp server stands in for a hosted
provider.

1. Start `llama-server` with a reasoning model and expose the
   OpenAI-compatible endpoint.

2. Configure the workbench server with `provider: "openai"` and set
   `openai_base_url` to the llama.cpp endpoint.

3. Set `llm.max_tokens` to a generous value, such as `8192`.

4. Open the estate overview.

The summary renders empty. The configured `llm.max_tokens` has no
bearing on the result, because the overview path uses the hardcoded
`512`.

## Expected behaviour

The analysis paths should honour the configured `llm.max_tokens`, and a
response that yields no text content should surface as an error rather
than an empty summary.

## Suggested fix

The following changes would address the problem:

- Make all three call sites read the configured `MaxTokens`, falling
  back to a sensible default when the setting is absent. The
  `(*Config).BuildClientOptions` helper added in `40426af` is already
  the shared choke point for the two server-side paths.

- Raise the default output budget for the summary paths so that a
  reasoning model has room for both its thinking block and a 3-5
  sentence answer.

- Treat a response with no text content as an error so the condition is
  reported rather than silently rendered as a blank summary.

- Document the setting's effect on the analysis paths in
  `docs/getting-started/configuration/server.md`.

## Notes

The LLM client code was reworked in `e831167` and `40426af`; the latter
is described in its own commit message as a behaviour-preserving
refactor, and it carried the constants through unchanged. The values
were last modified in `3d7a292`, which predates that work. No branch
currently on the remote changes them.

There is no `llamacpp` provider; the supported set is `anthropic`,
`openai`, `gemini`, and `ollama`. Local llama.cpp deployments should use
`provider: "openai"` with `openai_base_url`.

## Workaround

Until the fix lands, disable the model's thinking output so the 512-token
budget is enough for the summary. Either run a non-reasoning instruct
model, or start `llama-server` with `--jinja --reasoning-budget 0`, which
suppresses the thinking block for models that support it.

dpage:
Fixes #399

## Problem

Three AI-analysis paths hardcoded a small output-token cap that ignored the operator's configured `llm.max_tokens`:

| Path | Old value |
|---|---|
| Estate summary (`overview/generator.go`) | `llmMaxTokens = 512` |
| Server-info AI analysis (`api/server_info_handlers.go`) | `llmAnalysisMaxTokens = 512` |
| Alerter reasoning (`alerter/.../llm/reasoning.go`) | `pgllm.Int(500)` |

Reasoning models served through an OpenAI-compatible endpoint charge their thinking tokens against the same `max_tokens` budget as the answer. A 512-token budget is routinely consumed in full by the thinking block, so the model never emits a text block.

`extractTextFromResponse` collected only `pgllm.BlockText` blocks and returned an empty string, which callers treated as a successful generation. The estate summary therefore rendered blank with no error. This reproduces against a local llama.cpp server and was originally hit while building an all-in-one Docker stack.

The constants predate the `pgedge-go-llm-lib` migration and the `BuildClientOptions` extraction; both of those were behaviour-preserving refactors that carried the values through unchanged.

## Changes

- Resolve the output budget from configuration on all three paths, falling back to `4096` when the setting is unset, zero, or negative.
- `BuildClientOptions` now resolves max-tokens internally via the new nil-safe `AnalysisMaxTokens()` rather than accepting it as a parameter. A max-tokens *parameter* is exactly what let the two analysis call sites drift into hardcoding their own constants; removing it makes divergence impossible. Temperature stays a parameter because the two paths legitimately tune it.
- Treat a response carrying no usable text as an error (`ErrNoTextContent`) rather than an empty summary, so the condition surfaces instead of silently caching a blank panel. The error message names the likely cause and the setting to change.
- Map that condition to `502` on `GET /api/v1/server-info/{id}/ai-analysis`, documented in `openapi.go` and the regenerated `openapi.json`.
- Add `llm.max_tokens` to the alerter, which had no equivalent setting, with a `4096` default plus the config merge and defaulting passes.

## Note on the exported API

`BuildClientOptions` changes signature from `(maxTokens int, temperature float64)` to `(temperature float64)`. Both call sites are internal to this repository and are updated here.

## Behaviour deliberately left unchanged

The transient failure paths in `getAIAnalysis` (nil LLM config, no databases, client construction failure, chat error) still return `200` with a `null` body meaning "analysis unavailable". Only the "model answered but produced nothing usable" case becomes an error. Turning a provider outage into a `502` would reach well beyond this issue.

The web client's existing `.catch` in `ServerInfoDialog.tsx` already logs and leaves the panel empty, so no client change is needed.

## Testing

Table-driven tests cover each path for: configured value used; fallback on unset, zero, and negative; a small configured value still honoured; empty, whitespace-only, reasoning-block-only, and nil responses all yielding an error; non-empty responses passing through unchanged; and the `max_tokens` value actually observed on the wire, decoded from a stub server's request body.

Per-function line coverage on every touched function (verified with `go tool cover -func`):

```
llmproxy: AnalysisMaxTokens 100.0%  BuildClientOptions 100.0%
overview: extractTextFromResponse 100.0%  generateSummaryFromPrompt 100.0%  createLLMClient 100.0%
api:      getAIAnalysis 100.0%  writeAIAnalysisResponse 100.0%  createLLMClient 100.0%
          parseDatabaseAnalysisResponse 94.1%  handleServerInfoAI 91.3%
alerter:  Classify 100.0%  newLibReasoning 100.0%  NewReasoningProvider 100.0%
          SetLLMDefaults 100.0%  NewConfig 100.0%  LoadFromFile 100.0%
```

The single uncovered statement in `parseDatabaseAnalysisResponse` is in the pre-existing numbered-list-prefix stripper, untouched here.

`gofmt -l` is clean. `cd server && make test` and `cd alerter && make test` both pass. `go test ./internal/api/ -run OpenAPI -v` passes.

`make test-all` could not complete locally: the collector's lint step fails because the installed `golangci-lint` is built with Go 1.25 while the config targets Go 1.26.2. This reproduces on a pristine tree with all changes stashed, so it is pre-existing and unrelated to this change. `go vet ./...` is clean on both modules as a substitute; CI is the real gate.

https://claude.ai/code/session_01Piu2z26Jjg1fuPrkSJ5aFq

---
_Generated by [Claude Code](https://claude.ai/code/session_01Piu2z26Jjg1fuPrkSJ5aFq)_

## Summary by CodeRabbit

- **New Features**
  - Added configurable `llm.max_tokens` for AI analysis and alerter reasoning, defaulting to 4096 tokens.
  - Applied token limits across AI Overview, Ask Ellie, server analysis, and alert classification.
- **Bug Fixes**
  - Empty, whitespace-only, or reasoning-only model responses are now detected as unusable.
  - Server information analysis returns HTTP 502 with an actionable error when no usable analysis is produced; unavailable analysis remains `null`.
- **Documentation**
  - Updated configuration guides, API specifications, examples, and changelog with token-budget and error-handling details.

### DIFF
--- a/.claude/golang-expert/testing-strategy.md
+++ b/.claude/golang-expert/testing-strategy.md
@@ -136,6 +136,56 @@ one produces 401s or 403s that look like permission bugs.
 holding a named permission. The gate-ordering tests for mutating
 handlers are described in `rbac-patterns.md`.
 
+## HTTP Handler Tests: Never Point at Unreachable URLs
+
+The analysis paths build a real HTTP client through
+`(*llmproxy.Config).BuildClientOptions`, which sets
+`pgllm.Options.RequestTimeout` from `llm.timeout_seconds`; that setting
+defaults to 120 seconds in `server/src/internal/config/config.go`. A
+handler test that reaches a live provider endpoint, or an address that
+accepts the connection without answering, therefore blocks for up to two
+minutes instead of failing fast, and the cost lands on every `make test`
+and `make coverage` run. A locally refused connection is the benign
+case and returns at once.
+
+Stub the endpoint with `httptest.NewServer` and return a minimal valid
+body. The stubs in `server/src/internal/overview/generator_test.go` and
+`server/src/internal/api/server_info_handlers_test.go` are the patterns
+to copy; both decode the request body, which is also how the
+`max_tokens` assertions described below are made.
+
+When auditing a slow test, look for a base URL set to a literal
+`http://localhost:...` or to a real provider address.
+
+## AI Analysis Paths: Token Budget and Empty Responses
+
+The non-streaming analysis paths, the estate and scoped overview in
+`internal/overview/` and the server-info database analysis in
+`internal/api/server_info_handlers.go`, share two conventions that
+tests must uphold.
+
+The first convention is that no caller hardcodes an output-token
+budget. `(*llmproxy.Config).BuildClientOptions(temperature)` resolves
+the budget internally from `llm.max_tokens` through
+`AnalysisMaxTokens`, falling back to
+`llmproxy.DefaultAnalysisMaxTokens`, which is 4096, when the setting is
+unset, zero or negative. A caller that also stamps `MaxTokens` onto a
+`pgllm.ChatRequest` must read `AnalysisMaxTokens` rather than a local
+constant. Cover the configured, unset and negative cases, and assert
+the value on the wire by decoding `max_tokens` from the request body
+inside the `httptest` stub.
+
+The second convention is that a response carrying no text block, or
+whitespace-only text, is an error rather than an empty summary. Both
+paths return `llmproxy.ErrNoTextContent`, which the server-info handler
+maps to HTTP 502. Reasoning models charge their thinking blocks against
+the same budget as the answer, so this is the shape a starved budget
+produces; reproduce it with a content block of type `"thinking"`, for
+which the library exposes no named constant, or with an empty assistant
+message. The alerter mirrors both conventions in
+`alerter/src/internal/llm/reasoning.go`, using
+`config.DefaultLLMMaxTokens` and its own `ErrNoTextContent`.
+
 ## Linting
 
 Every sub-project runs golangci-lint v2 (`version: "2"` configs; note
--- a/alerter/src/internal/config/config.go
+++ b/alerter/src/internal/config/config.go
@@ -20,6 +20,19 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// DefaultLLMMaxTokens is the default output-token budget for tier-3
+// reasoning calls (llm.max_tokens).
+//
+// The previous hardcoded 500-token budget was sized for a classification
+// verdict alone. Reasoning models served through an OpenAI-compatible
+// endpoint (llama.cpp hosting Qwen3, DeepSeek-R1 distills, gpt-oss, and
+// similar) charge their thinking tokens against the same budget, so 500
+// tokens are routinely consumed in full by the thinking block and the
+// model never emits the verdict. 4096 leaves several thousand tokens of
+// thinking room ahead of the small JSON verdict, and matches the server's
+// llm.max_tokens default so operators see one number across the estate.
+const DefaultLLMMaxTokens = 4096
+
 // Config holds all configuration options for the alerter
 type Config struct {
 	// Datastore connection settings
@@ -147,13 +160,21 @@ type CorrelationConfig struct {
 
 // LLMConfig holds LLM provider settings
 type LLMConfig struct {
-	EmbeddingProvider string          `yaml:"embedding_provider"`
-	ReasoningProvider string          `yaml:"reasoning_provider"`
-	Ollama            OllamaConfig    `yaml:"ollama"`
-	OpenAI            OpenAIConfig    `yaml:"openai"`
-	Anthropic         AnthropicConfig `yaml:"anthropic"`
-	Voyage            VoyageConfig    `yaml:"voyage"`
-	Gemini            GeminiConfig    `yaml:"gemini"`
+	EmbeddingProvider string `yaml:"embedding_provider"`
+	ReasoningProvider string `yaml:"reasoning_provider"`
+
+	// MaxTokens caps the output tokens of a tier-3 reasoning call.
+	// Reasoning models charge their thinking blocks against this same
+	// budget, so a small value leaves no room for the classification
+	// verdict itself. A value of 0 or less selects
+	// llm.DefaultReasoningMaxTokens. Default: 4096.
+	MaxTokens int `yaml:"max_tokens"`
+
+	Ollama    OllamaConfig    `yaml:"ollama"`
+	OpenAI    OpenAIConfig    `yaml:"openai"`
+	Anthropic AnthropicConfig `yaml:"anthropic"`
+	Voyage    VoyageConfig    `yaml:"voyage"`
+	Gemini    GeminiConfig    `yaml:"gemini"`
 }
 
 // NotificationsConfig holds notification settings
@@ -340,6 +361,7 @@ func NewConfig() *Config {
 		LLM: LLMConfig{
 			EmbeddingProvider: "ollama",
 			ReasoningProvider: "ollama",
+			MaxTokens:         DefaultLLMMaxTokens,
 			Ollama: OllamaConfig{
 				BaseURL:        "http://localhost:11434",
 				EmbeddingModel: "nomic-embed-text",
@@ -374,8 +396,9 @@ func (c *Config) LoadFromFile(filename string) error {
 		return fmt.Errorf("failed to parse YAML config: %w", err)
 	}
 
-	// Apply defaults for notification settings
+	// Apply defaults for notification and LLM settings
 	c.SetNotificationDefaults()
+	c.SetLLMDefaults()
 
 	return nil
 }
@@ -512,6 +535,16 @@ func GetDefaultConfigPath(binaryPath string) string {
 // ConfigFileExists checks if a config file exists at the given path
 func ConfigFileExists(path string) bool {
 	return fileutil.FileExists(path)
+}
+
+// SetLLMDefaults sets default values for LLM config that a YAML file may
+// have left unset. NewConfig already seeds these defaults, but a caller
+// that loads into a bare Config (or a YAML file that sets llm.max_tokens
+// to 0) must still end up with a usable budget.
+func (c *Config) SetLLMDefaults() {
+	if c.LLM.MaxTokens <= 0 {
+		c.LLM.MaxTokens = DefaultLLMMaxTokens
+	}
 }
 
 // SetNotificationDefaults sets default values for notification config
--- a/alerter/src/internal/config/config_test.go
+++ b/alerter/src/internal/config/config_test.go
@@ -633,7 +633,7 @@ llm:
 			want int
 		}{
 			{
-				name: "explicit value is honoured",
+				name: "explicit value is honored",
 				yaml: "llm:\n  max_tokens: 16384\n",
 				want: 16384,
 			},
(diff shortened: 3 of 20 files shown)


AntTheLimey (changes requested):
One fix before this lands. `.claude/golang-expert/testing-strategy.md`
re-inflates the file you rewrote this morning in #444, from 169 lines
back to 1151; the appended 982 lines are almost entirely the pre-trim
document (directory-structure diagram, generic table-driven example, a
"Testing Stack" section that contradicts the rewritten file's own
"manual mocks, no external framework" note, the "Common Pitfalls"
list). That rewrite was 163 insertions against 941 deletions, and its
opening paragraph sets the policy the re-inflation breaks: the file
"deliberately omits general Go testing advice", and where it disagrees
with a Makefile, a workflow or the code, the code wins. Two short
passages in the same diff are new and worth keeping:
"HTTP Handler Tests: Never Point at Unreachable URLs" and "AI Analysis
Paths: Token Budget and Empty Responses". Please rebase that file's
edit onto the current 169-line version and land only those two
sections.

Everything else here is in good shape. `AnalysisMaxTokens` and the
config floor correctly turn an unset, zero, or negative `max_tokens`
into the 4096 default, and a small-but-positive value (the case that
would previously go silently blank) is now surfaced as an explicit
`ErrNoTextContent` / HTTP 502 rather than swallowed — I traced this
through the alerter's fail-safe-to-alert path, the estate overview's
refresh (which now keeps serving the last good summary instead of
overwriting it with a blank one), and the server-info handler's 502
mapping, and confirmed with tests that decode `max_tokens` off the wire
that the configured value actually reaches the provider request on all
three paths, not just the config struct. `gofmt`, both sub-projects'
`make lint`, and `go test ./internal/api/ -run OpenAPI` are all clean;
`make openapi` regenerates byte-identical to what's committed; server
and alerter coverage both come in at or above their ratchets and above
main (56.3%/70.0% vs 56.1%/69.9%), with every touched function at or
near 100%, matching the numbers in your PR description. No CodeRabbit
or human findings are outstanding (CodeRabbit hit its rate limit and
posted no content; Codacy reports 0 issues).

On merge order: this conflicts textually with #419 only in
`testing-strategy.md` (both append to the same base) — no code overlap,
since #419 touches CLI flag plumbing, not `LLMConfig` or `internal/llm/`.
It also conflicts with #386 and #396 only in adjacent
`docs/changelog.md` entries. All three are ordinary merge conflicts for
whoever lands second, not blockers.

dpage:
Thanks — the re-inflation was a real defect and you were right about the cause. This branch predates #444, and the rebase carried the pre-trim document forward under a second top-level heading without conflicting, which is exactly why it went unnoticed.

`.claude/golang-expert/testing-strategy.md` is now the 171-line version from `main` plus the two sections, at 221 lines. The net change against `main` is +50 lines rather than +982. The fix is squashed into the commit that introduced it, so there is no add-then-remove churn in the history. None of the four items you named are present, and exactly one top-level heading remains.

## One deviation, which needs your call

I did not re-add `HTTP Handler Tests: Never Point at Unreachable URLs` verbatim, because it does not describe the current code. Checking each of its claims against this branch:

- `defaultLLMHTTPTimeout` does not exist anywhere under `server/src/`.
- `TestHandleModels_OpenAIBaseURLOnly` does not exist; it is cited as the canonical pattern.
- `HandleModels` does not exist anywhere under `server/src/`.
- `httptest.NewServer` appears zero times in `internal/llmproxy/proxy_test.go`.
- `internal/chat/llm.go` issues no outbound HTTP at all; it contains `BuildSystemPrompt`, `BuildUserContext` and `sanitizeMemoryField`.

So the passage reads as new against `main` only because #444 deleted it. Re-adding it as written would have reintroduced precisely the rot that rewrite removed, and would have contradicted the file's own opening policy that the code wins.

The underlying principle is still sound, so I kept the section and rewrote it against facts I verified:

- The 120-second budget is real, but it is `llm.timeout_seconds` (`server/src/internal/config/config.go`, default 120), applied as `pgllm.Options.RequestTimeout` in `internal/llmproxy/proxy.go`.
- The working `httptest` stubs are in `internal/overview/generator_test.go` and `internal/api/server_info_handlers_test.go`; both decode the request body, which is how the `max_tokens` assertions are made.
- I corrected the failure mode: a locally refused connection returns immediately. The expensive case is a live provider endpoint, or an address that accepts the connection without answering.

`AI Analysis Paths: Token Budget and Empty Responses` needed no correction. I verified `AnalysisMaxTokens`, `DefaultAnalysisMaxTokens` at 4096, the unset/zero/negative fallback, `ErrNoTextContent`, the 502 mapping, and the alerter's `config.DefaultLLMMaxTokens` mirror against the code.

If you would rather the first section were simply dropped than rewritten, say so and I will remove it.

## Also dropped

The diff contained a third new section, `Handler Routing Tests Without a Live Datastore`, documenting the lazily-created `pgxpool` aimed at a closed port used by `TestHandleServerInfoAI`. You asked for only the two named sections, so I dropped it. Flagging it in case that was an oversight rather than a decision, since it documents a pattern this PR actually introduces.

## Rebase

Rebased onto `a80a294`, which was 48 commits ahead. One conflict, in `docs/changelog.md`. It needed more than a both-sides resolution: my entry is a Fixed item, and the three-way merge landed it inside a `### Removed` block newly inserted on `main`. I relocated it to the end of `### Fixed` and left your `### Removed` section as written.

Unrelated, and left alone: `main`'s Unreleased section already carries two separate `### Removed` headings. That predates this branch, so folding it in here would only widen the diff.

`gofmt` is clean and the server and alerter suites both pass against the new base. Thank you for running both sub-projects' `make lint` — that is the one gate this container cannot run, because its `golangci-lint` is built with Go 1.25 against a config targeting 1.26.2.

---
_Generated by [Claude Code](https://claude.ai/code)_

dpage:
Both CodeRabbit findings verified against the code and fixed. Pushed as `5b5428a`.

**Unparseable analysis cached as success** (`server_info_handlers.go`) — this one was a real bug in my change, and worth calling out. `parseDatabaseAnalysisResponse` rejected empty and whitespace-only text, but text that was non-empty yet matched no known database fell through as an empty map with a nil error. `getAIAnalysis` then cached that for the full TTL and the handler served it as a successful blank analysis. That is the same failure mode this PR exists to remove, and the comment on `getAIAnalysis` already claimed such responses were reported rather than cached, so the code contradicted its own stated intent. An empty parse is now `ErrNoTextContent`.

I checked the guard order before making an empty result an error: `getAIAnalysis` returns on an empty `databases` slice at line 708, well before the parse at line 742, so "nothing to analyse" still yields `200` with a null body and can never become a `502`. That precondition is now recorded on the parser's doc comment, since the new branch depends on it.

Two tests added, no existing test modified — none had asserted the old behaviour. `parseDatabaseAnalysisResponse` moved from 94.1% to 97.2% line coverage; `getAIAnalysis` and `writeAIAnalysisResponse` are at 100%. `openapi.go` needed no change: the `502` entry for this endpoint already reads "The model returned no usable analysis text", which describes the new branch exactly.

**Alerter example wording** (`examples/ai-dba-alerter.yaml`) — correct, the comment understated the setting's reach. `Classify` is called from both `engine/anomalies.go:416` and `engine/reevaluation.go:105`, so the comment now names anomaly candidate classification and acknowledged-alert re-evaluation. `docs/getting-started/configuration/alerter.md` already described it as "per reasoning call", so no change was needed there.

On the docstring-coverage pre-merge warning, I have deliberately not acted, and would rather say so than leave it looking overlooked. All six non-test functions this PR adds carry doc comments. The undocumented functions it counts are eleven pre-existing helpers in `openapi.go` (`jsonResponse`, `pathParamInt` and similar), which are undocumented on `main` and which this PR touches only via a two-line `502` entry, plus seventeen test functions, which do not carry doc comments by convention here. Documenting either group to clear a threshold would mean churn in unrelated code. Happy to do it if you would rather the check were green.

`gofmt` clean; `cd server && make test` passes with no `FAIL`. `make coverage` cannot meet `COVERAGE_MIN` in my container because it has no PostgreSQL, so every database integration test skips; that is environmental and CI runs against real PG service containers.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #413, opened 2026-08-12
Honour llm.max_tokens on the AI analysis paths

### LINKED ISSUE #399: AI analysis paths hardcode a 512-token output cap, producing empty summaries with local reasoning models

dpage:
## Summary

The estate summary, the Server Info database analysis, and the alerter's
tier 3 classification each hardcoded a 512-token output budget, so the
configured `llm.max_tokens` had no bearing on any of them. A reasoning
model spends part of its budget on an internal thinking block before it
emits any answer text, so with a local model served over an
OpenAI-compatible endpoint the cap was consumed entirely by reasoning;
the response then carried no text block, and because the callers treated
the resulting empty string as a successful generation the failure
surfaced as a blank panel with no error at all.

The changes are as follows:

- Add `(*llmproxy.Config).AnalysisMaxTokens`, which reports the
  operator-configured `llm.max_tokens` and falls back to the new
  `DefaultAnalysisMaxTokens` (4096) when the setting is absent or
  non-positive, and use it for both the estate summary and the Server
  Info analysis, on the chat request and on the client options alike.

- Give the alerter an `llm.max_tokens` setting, which it previously
  lacked entirely, with the same 4096 default and fallback rule, and
  plumb it through `newLibReasoning` into the classification request.

- Report a response that yields no usable text rather than rendering it:
  the overview path returns an error naming the configured budget, the
  Server Info path logs and skips the cache so the next request retries
  instead of serving a blank result, and the alerter's `Classify`
  returns an error, which its caller already handles by failing safe to
  an alert.

- Document the setting's effect on the analysis paths in the server and
  alerter configuration guides, update both commented example
  configurations, and add a changelog entry.

## Test plan

- New unit tests cover `AnalysisMaxTokens` and `ReasoningMaxTokens`
  across nil, unset, zero, negative, and configured values.

- New table-driven tests drive the overview and Server Info paths
  against an OpenAI-compatible stub that decodes the request body, and
  assert that the budget on the wire is the configured value, or the
  4096 default when the setting is unset.

- New tests assert that a response carrying no text (empty content, and
  whitespace only) produces an error naming the configured budget, and
  that the Server Info path leaves the cache untouched in that case.

- Every changed function reports 100% line coverage:
  `AnalysisMaxTokens`, `generateSummaryFromPrompt`, both
  `createLLMClient` methods, `getAIAnalysis`, `ReasoningMaxTokens`,
  `Classify`, and `newLibReasoning`.

- `make lint` passes with zero issues for both `server` and `alerter`,
  and `gofmt` reports no changes.

- `make test-all` passes bar one pre-existing failure in
  `server/internal/tools` (`TestStoreMemoryGeneratesEmbeddingIntegration`
  and `TestRecallMemoriesGeneratesQueryEmbeddingIntegration`), which is
  a stale `vector(3)` fixture against the 4000-dimension padding and
  reproduces identically on a clean checkout of `main`; it is unrelated
  to this change and skips in CI, where Postgres lacks pgvector.

Closes #399

## Summary by CodeRabbit

* **New Features**
  * Added configurable LLM response token limits for chat, database analysis, estate summaries, and alert classification.
  * Added a default 4,096-token budget when limits are unset or invalid.
  * Reasoning models now account for their reasoning budget within the configured limit.

* **Bug Fixes**
  * Empty or unusable AI responses are reported as errors and are not cached.

* **Documentation**
  * Updated configuration guides, example files, and the changelog with token-limit behavior and settings.

### DIFF
--- a/alerter/src/internal/config/config.go
+++ b/alerter/src/internal/config/config.go
@@ -20,6 +20,12 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+// DefaultLLMMaxTokens is the output-token budget applied to reasoning
+// requests when llm.max_tokens is unset or non-positive. It is generous
+// enough that a reasoning model can emit a thinking block and still have
+// room for the classification that follows it.
+const DefaultLLMMaxTokens = 4096
+
 // Config holds all configuration options for the alerter
 type Config struct {
 	// Datastore connection settings
@@ -147,13 +153,31 @@ type CorrelationConfig struct {
 
 // LLMConfig holds LLM provider settings
 type LLMConfig struct {
-	EmbeddingProvider string          `yaml:"embedding_provider"`
-	ReasoningProvider string          `yaml:"reasoning_provider"`
-	Ollama            OllamaConfig    `yaml:"ollama"`
-	OpenAI            OpenAIConfig    `yaml:"openai"`
-	Anthropic         AnthropicConfig `yaml:"anthropic"`
-	Voyage            VoyageConfig    `yaml:"voyage"`
-	Gemini            GeminiConfig    `yaml:"gemini"`
+	EmbeddingProvider string `yaml:"embedding_provider"`
+	ReasoningProvider string `yaml:"reasoning_provider"`
+
+	// MaxTokens is the output-token budget for a reasoning request.
+	// A reasoning model spends part of that budget on its thinking block
+	// before it emits any answer, so a tight cap can leave no room for
+	// the classification itself. Defaults to DefaultLLMMaxTokens when
+	// unset or non-positive.
+	MaxTokens int `yaml:"max_tokens"`
+
+	Ollama    OllamaConfig    `yaml:"ollama"`
+	OpenAI    OpenAIConfig    `yaml:"openai"`
+	Anthropic AnthropicConfig `yaml:"anthropic"`
+	Voyage    VoyageConfig    `yaml:"voyage"`
+	Gemini    GeminiConfig    `yaml:"gemini"`
+}
+
+// ReasoningMaxTokens returns the output-token budget for a reasoning
+// request, falling back to DefaultLLMMaxTokens when the operator has not
+// configured a positive llm.max_tokens.
+func (l *LLMConfig) ReasoningMaxTokens() int {
+	if l == nil || l.MaxTokens <= 0 {
+		return DefaultLLMMaxTokens
+	}
+	return l.MaxTokens
 }
 
 // NotificationsConfig holds notification settings
@@ -340,6 +364,7 @@ func NewConfig() *Config {
 		LLM: LLMConfig{
 			EmbeddingProvider: "ollama",
 			ReasoningProvider: "ollama",
+			MaxTokens:         DefaultLLMMaxTokens,
 			Ollama: OllamaConfig{
 				BaseURL:        "http://localhost:11434",
 				EmbeddingModel: "nomic-embed-text",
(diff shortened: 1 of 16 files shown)


dpage:
Closing as a duplicate. PR #411 fixes the same three hardcoded output-token caps and goes further, also updating `openapi.go`, the generated `openapi.json`, the server-info documentation and the golang-expert knowledge base. There is nothing in this branch that #411 does not already cover, so #411 is the one to review for #399.

PR #414, opened 2026-08-12
Explain why utility statements have no query plan

### LINKED ISSUE #368: Query Plan panel shows raw syntax error for non-explainable statements (VACUUM, ANALYZE, etc.)

dpage:
## Summary

The Query Plan panel built its request by concatenating `EXPLAIN` with
the captured statement text, and because `pg_stat_statements` records
utility and DDL commands alongside DML, a Top Queries row can hold bare
`VACUUM`, `ANALYZE`, `REINDEX ...` or `CLUSTER ...` text; wrapping any
of those in `EXPLAIN` is invalid SQL, so the panel surfaced a raw
`syntax error at or near "VACUUM"` back at the user.

This change adds `client/src/utils/sqlHelpers.ts`, which classifies a
statement from its leading keyword after skipping leading whitespace,
line and block comments, and opening parentheses. `useQueryPlan` now
checks that classification before building the request and reports the
statement type through a new `notExplainable` field instead of firing a
request that can only fail, whilst `QueryPlanPanel` renders a friendly
informational alert naming the statement type in place of the Postgres
error.

The explainable set is SELECT, INSERT, UPDATE, DELETE, MERGE, VALUES,
EXECUTE, DECLARE, WITH, TABLE, CREATE TABLE AS and CREATE MATERIALIZED
VIEW, plus REFRESH MATERIALIZED VIEW, which the EXPLAIN grammar accepts
even though the documentation omits it. Every case in the test tables
was verified empirically against PostgreSQL 18.4 rather than assumed.

## Test plan

- New `sqlHelpers` unit tests cover keyword extraction, comment and
  parenthesis stripping, and eighteen explainable and twenty-three
  non-explainable statements; the file reports 100% line, branch and
  function coverage.
- New `useQueryPlan` tests assert that VACUUM, ANALYZE, REINDEX,
  CLUSTER, CREATE INDEX and TRUNCATE issue no API call at all, that the
  detected keyword is reported, and that the flag clears when a
  plannable query follows; the hook reports 98.68% line coverage.
- New `QueryPlanPanel` tests assert the friendly message renders, that
  the plan tabs stay hidden, and that the notice takes precedence over
  a stale error; the component reports 100% line coverage.
- `npm run lint` reports 0 errors, and `npm run build` succeeds.

Closes #368

### DIFF
--- a/client/src/components/Dashboard/ObjectDashboard/QueryPlanPanel.tsx
+++ b/client/src/components/Dashboard/ObjectDashboard/QueryPlanPanel.tsx
@@ -22,6 +22,7 @@ import { Refresh as RefreshIcon } from '@mui/icons-material';
 import CollapsibleSection from '../CollapsibleSection';
 import PlanTree from './PlanTree';
 import { useQueryPlan } from '../../../hooks/useQueryPlan';
+import { buildNotExplainableMessage } from '../../../utils/sqlHelpers';
 import { spinKeyframes } from '../styles';
 
 interface QueryPlanPanelProps {
@@ -93,6 +94,7 @@ const QueryPlanPanel: React.FC<QueryPlanPanelProps> = ({
         jsonPlan,
         loading,
         error,
+        notExplainable,
         fetch: fetchPlan,
     } = useQueryPlan(queryText, connectionId, databaseName);
 
@@ -149,7 +151,13 @@ const QueryPlanPanel: React.FC<QueryPlanPanelProps> = ({
                 </Box>
             )}
 
-            {error && !textPlan && !jsonPlan && (
+            {notExplainable !== null && (
+                <Alert severity="info" sx={{ mt: 1 }}>
+                    {buildNotExplainableMessage(notExplainable)}
+                </Alert>
+            )}
+
+            {notExplainable === null && error && !textPlan && !jsonPlan && (
                 <Alert severity="info" sx={{ mt: 1 }}>
                     {isParameterError(error)
                         ? PARAM_ERROR_MSG
--- a/client/src/components/Dashboard/ObjectDashboard/__tests__/QueryPlanPanel.test.tsx
+++ b/client/src/components/Dashboard/ObjectDashboard/__tests__/QueryPlanPanel.test.tsx
@@ -52,6 +52,7 @@ describe('QueryPlanPanel', () => {
             jsonPlan: null,
             loading: false,
             error: null,
+            notExplainable: null,
             fetch: mockFetch,
         });
     });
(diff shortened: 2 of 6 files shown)


dpage:
Closing in favour of #385, which fixes the same issue more correctly.

Verified against PostgreSQL 18.4: `EXPLAIN REFRESH MATERIALIZED VIEW mv` is accepted by the parser and returns "Utility statements have no plan structure", with `FORMAT JSON` returning `["Utility Statement"]`. This branch classifies that statement as explainable, so it still issues the request and then hands the plan renderer a JSON array containing a bare string, replacing one failure mode with another. #385 models three states (plan, planless, unsupported) and hardens `extractJsonPlan` against non-plan shapes, so it handles the case correctly; its CTAS pattern is also tighter, and it updates the changelog.

The `sqlHelpers` extraction and the dedicated `notExplainable` field here are a reasonable follow-up on top of a merged #385, but are not a reason to keep this branch open.

PR #420, opened 2026-08-12
Standardise deferred pgx Rollback() calls on a non-cancelable context

### LINKED ISSUE #381: Standardise deferred pgx Rollback() calls on a non-cancelable context
Deferred `tx.Rollback()` calls are inconsistent across the tree: most pass the request-derived `ctx`, whilst two sites deliberately pass `context.Background()`. The hardened sites exist because a cancelled context can trigger the pgx v5 close-of-closed-channel panic described in [jackc/pgx#2470](https://github.com/jackc/pgx/issues/2470), which leaves the pooled connection leaked in an aborted-transaction state; the rationale is documented at `server/src/internal/database/connection_queries.go:442` and again in `handleTopQueries` in `server/src/internal/api/perf_summary_handlers.go`.

The remaining sites should be converted in one deliberate sweep, with the rationale documented once (ideally in the golang-expert knowledge base rather than repeated at each call site) and a short note in the contributor guidance so that new transactions get it right by default.

Sites still passing the request context:

- `server/src/internal/api/perf_summary_handlers.go` - `handlePerfSummary` (~line 264), `handleDatabaseSummaries` (~line 743)
- `server/src/internal/database/cluster_queries.go` - ~lines 470, 530
- `server/src/internal/database/relationship_queries.go` - ~lines 182, 239, 391
- `server/src/internal/database/alert_queries.go` - ~lines 383, 445 (these two also lack the `//nolint:errcheck` annotation the others carry)
- `alerter/src/internal/database/alert_queries.go` - ~line 267

Already hardened, for reference: `server/src/internal/database/connection_queries.go:442` and `handleTopQueries`. `server/src/internal/auth/token_scope.go:457` uses `database/sql` and takes no context, so it needs no change.

Raised as a follow-up from the review of #379.

dpage:
## Summary

Deferred `tx.Rollback()` calls were inconsistent across the tree: two
sites deliberately passed `context.Background()`, each carrying its own
copy of the explanation, whilst the remaining twenty passed the
request-derived `ctx`. This sweeps every rollback in the server,
collector and alerter through a new shared helper,
`pkg/rollback`, in one deliberate pass.

The reason it matters is that pgx v5 treats a rollback on a cancelled
context as a failed rollback and calls `conn.die()`, so a client that
closed a tab or timed out mid-request cost the pool a connection whilst
it redialled (connection churn rather than a leak: the server tears the
transaction down when the socket closes). A bare `context.Background()`
fixes that but has no deadline, and pgconn skips its context watcher
for it, so a rollback to a wedged peer could pin a pool slot and the
goroutine indefinitely. The helper therefore runs the rollback on
`context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)`,
keeping the request's tracing values and bounding the wait.

Transaction semantics, commit paths and error handling are untouched;
only the context handed to the rollback changes.

### The helper (`pkg/rollback`)

- `rollback.Tx(ctx, tx)` for a plain rollback, safe to defer directly
  since the timeout starts when the deferred call runs.
- `rollback.ToSavepoint(ctx, tx, name)` for `ROLLBACK TO SAVEPOINT`,
  used by `runSavepointed` in the collector.
- `rollback.Context(ctx)` for anything else.
- `rollback.Scan(root)` backs a convention test in each of the four Go
  modules (server, collector, alerter, pkg) that fails on any direct
  single-argument `Rollback` call or hand-written `ROLLBACK` SQL outside
  the package, so the rule runs under each module's own path-filtered
  CI. The rationale lives once in the package doc comment.

### Sites converted

- `server/src/internal/api/perf_summary_handlers.go`: `handlePerfSummary`,
  `handleDatabaseSummaries`, `handleTopQueries`
- `server/src/internal/api/perf_summary_connection_groups.go`
- `server/src/internal/api/query_handlers.go`: the deferred cleanup in
  `executeQuery`
- `server/src/internal/database/cluster_queries.go`,
  `relationship_queries.go`, `alert_queries.go`, `connection_queries.go`
- `server/src/internal/tools/transaction.go`: seven rollbacks across
  `BeginReadOnlyTx` and `BeginTx`
- `alerter/src/internal/database/alert_queries.go`: `ReactivateAlert`
- `collector/src/probes/storage.go`: `StoreMetrics`
- `collector/src/database/schema.go`: both `Migrate` error paths and
  `runSavepointed`

`server/src/internal/auth/` uses `database/sql`, whose `Rollback()`
takes no context, so it is out of scope.

### Documentation

- `.claude/golang-expert/transaction-rollback.md` describes the helper
  and the convention test for the golang-expert agent.
- A "Transaction Rollbacks" note under "Go Code" in
  `docs/developer-guide/contributing.md`.
- `docs/changelog.md` records the fix.

## Test plan

- `pkg/rollback` unit tests prove the rollback context is not cancelled
  when the parent is, carries the parent's values, has a deadline within
  `Timeout`, and that `ToSavepoint` issues the right SQL and rejects
  non-identifier names.
- `server/src/internal/database/rollback_context_integration_test.go`
  drives `UnacknowledgeAlert` over a single-connection pool with a pgx
  `QueryTracer` that cancels the request context the moment the last
  statement completes, then asserts the `ROLLBACK` ran on an uncancelled
  context with a deadline and that the pooled backend PID is unchanged.
- `server/src/internal/tools/transaction_rollback_integration_test.go`
  covers all seven rollbacks in `BeginReadOnlyTx`/`BeginTx`.
- Each module's `TestNoDirectRollbacksInModule` was verified to fail when
  one site is reverted to a direct call.
- Handler tests drive the three performance-summary endpoints and the
  query executor end to end against Postgres over a trimmed metrics
  schema.

Every function touched is at or above the 90% line-coverage floor;
`pkg/rollback` is at 95.9%. `gofmt`, `go vet` and `golangci-lint` are
clean in all four modules.

### Known gaps

- `pkg/` has no path-filtered CI workflow of its own, so a pkg-only PR
  runs only `ci-docker.yml`; the root `Makefile` does test it. Worth a
  separate issue.
- The rollback in `StoreMetrics` is unreachable because both failure
  paths shadow the outer `err` with `:=`; pre-existing, left for its own
  issue.

### Merge order

This branch sits on #386's current head, so the PR shows #386's commit
until it merges; a plain rebase onto `main` will then be a no-op.

Closes #381

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database transaction cleanup across server, collector, and alerter workflows.
  * Rollbacks now complete reliably even when requests are canceled, helping preserve connection-pool health.
  * Savepoint rollback handling now uses bounded operations and validates savepoint names.

* **Documentation**
  * Added contributor guidance and changelog information for transaction rollback behavior.

* **Tests**
  * Added extensive coverage for rollback failures, canceled requests, migration errors, connection reuse, and transaction-related API workflows.
  * Added checks to prevent unsupported direct rollback implementations.


dpage:
@coderabbitai review

AntTheLimey (changes requested):
Two things to change before this merges.

The first is a merge hazard rather than a defect in the change itself.
`query_execute_rollback_test.go` calls `NewConnectionHandler` three
times, and #396 deletes that constructor. The two PRs touch disjoint
files so git merges them without complaint, but I built the merged tree
with `git merge-tree` and `go vet` fails on it with "undefined:
NewConnectionHandler", in either merge order, and the branch protection
on main will not catch it because Codacy is the only required check.
Could you switch those three calls to `NewConnectionHandlerWithSecurity(
ds, store, checker, false, nil, nil)`? It is exactly equivalent —
`NewConnectionHandler` sets `DefaultHostValidator()`, which is defined as
`NewHostValidator(false, nil, nil)` — and it compiles against today's
main and against main after #396, which makes the ordering irrelevant. I
confirmed both, and that the executeQuery tests still pass with the
substitution.

The second is about the construction itself. `context.Background()` is
non-cancelable, but it is also deadline-free, and pgconn special-cases it
by identity: `PgConn.Exec` skips installing its context watcher when the
context is literally `context.Background()`, so no socket deadline is
ever set for the ROLLBACK. `pgxpool.Tx.Rollback` releases the connection
only after the inner rollback returns, so a peer that ACKs but never
answers — paused PgBouncer, wedged backend, a firewall holding the flow
open — pins a pool slot and the handler goroutine with nothing to break
it but TCP keep-alive. That is a change in behaviour, not just in
spelling: all four HTTP sites derived their context with a 30s timeout,
so the rollback used to be bounded at 30 seconds. With `pool_max_conns`
defaulting to 4, four of these exhaust a monitored connection's pool, and
`http.Server.Shutdown` waits on the goroutines. A bounded non-cancelable
context gets you the fix without the new failure mode — a small helper
returning `context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)`
would do it, and `WithoutCancel` keeps the tracing values `Background()`
throws away. Worth settling now because the convention test asserts the
literal `context.Background()`, so the bounded form currently fails it.

Three smaller things, take or leave. The convention test matches a
spelling rather than a property, and all three gaps are live: it does not
see `runSavepointed`'s `tx.Exec(ctx, "ROLLBACK TO SAVEPOINT ...")` in the
collector, it does not scan `pkg/`, and it rejects
`context.WithoutCancel(ctx)` even though the failure message asks for a
non-cancelable context. Relatedly the test only runs in the server
module, and `ci-server.yml` is path-filtered to `server/**`, so a
collector-only or alerter-only PR never runs the rule that is supposed to
cover all three. And I think the changelog overstates the symptom: I
reverted the two `alert_queries.go` rollbacks and re-ran your integration
test, and the backend PID moved 79 → 80 with no `idle in transaction`
rows left anywhere — `conn.die()` closes the socket and the server tears
the transaction down, so the cost is connection churn rather than a leak.
In the same vein, jackc/pgx#2470, cited in the convention test's doc
comment, is closed and was filed against v5.7.0, while this repository
pins v5.9.2; it is thinner support for the convention than the comment
implies.

Everything else measured clean. I reproduced the defect by reverting the
two alert-queries rollbacks, watched `TestDeferredRollbackIgnoresCanceled
RequestContext` fail on both the cancelled-context assertion and the
backend-PID assertion, and watched it pass again once restored — the
tests genuinely exercise a cancelled parent and prove the connection
survived, which is more than most regression tests of this shape manage.
Every pgx rollback in all four modules is converted, with no site missed;
the six `database/sql` rollbacks in `internal/auth` are correctly out of
scope. gofmt and golangci-lint are clean in all three sub-projects, `go
vet` passes, and every function you touched is at or above the 90% floor
(lowest is `handleDatabaseSummaries` at 90.0%, highest `BeginTx` and
`BeginReadOnlyTx` at 100%). Server total is 58.2% against 56.1% on main,
alerter 70.0% against 69.9%. The two red collector jobs are not yours:
collector measures 86.0% on the branch against 85.9% on main locally,
and CI fails only on PG 14 and 15, where the log reads 85.4% against the
86 ratchet — that ratchet is unreachable on the older matrix entries
because version-gated tests skip there, which is a #444 question rather
than a #420 one. No OpenAPI surface changed, the
changelog entry is in the right section, and the new knowledge-base file
duplicates nothing in `testing-strategy.md`.

On merge order: #396 as above. #386 also touches
`perf_summary_handlers.go` and auto-merges, but it rewrites the
top-queries CTE to select `userid`, `min_exec_time` and `max_exec_time`
and to join `pg_stat_activity` on `usesysid`/`usename`, none of which
exist in your new `perfEndpointTestSchema` fixture — the merged tree
compiles and then fails in the two top-queries tests, so whoever lands
second extends the fixture. #442 and #386 both conflict with you in
`docs/changelog.md`, trivially. #411 merges clean.

dpage:
Rebased for the merge-order interactions only; Ant's review findings (the bounded non-cancelable context helper, the convention test's gaps and the changelog wording) remain open and are not addressed here.

- The branch now sits on #386's head (4fe39024), since that lands first, so this PR shows #386's commit until it merges; after that a plain rebase onto `main` will be a no-op.
- The three `NewConnectionHandler` calls in `query_execute_rollback_test.go` are replaced with the equivalent `NewConnectionHandlerWithSecurity(ds, store, checker, false, nil, nil)`, so the tree compiles against `main` after #396.
- `perfEndpointTestSchema` gains the columns #386's top-queries CTE selects (`userid`, `min_exec_time`, `max_exec_time` on `pg_stat_statements`; `usesysid`, `usename` on `pg_stat_activity`), the seed exercises both a resolved and an unresolved role, and the snapshot test asserts the username and min/max times. The `invalid order_by` case now compares against the generated whitelist message rather than a hardcoded column list.

Gates on the stacked tree: gofmt, vet and `golangci-lint` clean in server, collector and alerter; full suites with `-race` pass at 59.3% (server), 88.1% (collector) and 70.0% (alerter).

dpage:
Thanks Ant, all of that is addressed in 5f9ae62e.

- The rollback context is now bounded as you suggested: a new shared package `pkg/rollback` runs every rollback on `context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)`, via `rollback.Tx(ctx, tx)` for plain rollbacks and `rollback.ToSavepoint(ctx, tx, name)` for `runSavepointed`'s `ROLLBACK TO SAVEPOINT`. The timeout starts when the deferred call runs, not at `Begin`.
- The convention test now checks the property rather than the spelling: `rollback.Scan` flags any direct single-argument `Rollback` call and any hand-written `ROLLBACK` SQL outside the package, and a small test in each of server, collector, alerter and pkg scans its own module root, so it runs under each module's path-filtered CI. It catches the savepoint case, and I confirmed it fails when one site is reverted.
- The changelog, contributor note, knowledge-base file and doc comments now describe the cost as connection churn and a lost pool slot whilst the pool redials, not a leak, and jackc/pgx#2470 is mentioned only as related history in the package doc.

Two things I have not changed: `pkg/` has no path-filtered workflow of its own (a pkg-only PR triggers only `ci-docker.yml`), and the unreachable rollback in `StoreMetrics` where both failure paths shadow `err`; both are noted in the description as candidates for their own issues.

The branch is rebased onto #386's current head (baa9704b), so the PR still shows that commit until #386 lands.

dpage:
Codacy: the Opengrep concat-sqli finding on `pkg/rollback/rollback.go:99` (`"ROLLBACK TO SAVEPOINT "+name`) is marked as a false positive. The name is validated as a plain identifier (`^[A-Za-z_][A-Za-z0-9_]*$`) before any SQL is built, and the statement takes no bind parameters, so there is no other way to issue it. The remaining Lizard complexity warning is in #386's `useQueryStats.ts` and disappears once #386 merges.

dpage:
Rebased onto main now that #386 (and #449, #450) have landed, so the PR shows only its own commits. Two expectations in the performance-endpoint rollback test needed following #450: the cache hit ratio is now a nullable per-interval figure (92.6% for the fixture's two samples rather than the absolute 90%) and the invalid time-range message comes from the shared resolver. Full server, collector and alerter suites and the per-module convention tests pass locally.

PR #433, opened 2026-08-27
fix(metrics): build the bucket series in FROM, not the select list

vyruss:
## Summary

The gap-filling bucket series in the metrics time-series queries is
built with `generate_series` in the select list. This moves it into the
`FROM` clause. The two forms return identical rows in PostgreSQL, so
this is a pure rewrite of the query text with no behaviour change.

The reason for the change is that a set-returning function in a select
list has no equivalent in DuckDB. When the `metrics.*` tables are read
through a view whose query is executed by DuckDB rather than
PostgreSQL, DuckDB's timestamp `generate_series` returns a LIST in that
position instead of a set, and the read fails with:

```
Conversion Error: Unimplemented type for cast
(TIMESTAMP WITH TIME ZONE -> TIMESTAMP WITH TIME ZONE[])
```

It is the placement that matters, not the arguments; the failure
reproduces with all-literal arguments. Moving the call into `FROM`
works on both backends, so the change is applied unconditionally
rather than behind a flag.

## Changes

Both sites are the `all_buckets` CTE, in `BuildMetricsQuery` and
`BuildDerivedMetricsQuery`.

Before:

```sql
all_buckets AS (
    SELECT generate_series($3::timestamptz, $4::timestamptz, $1::interval) AS bucket_time
)
```

After:

```sql
all_buckets AS (
    SELECT bucket_time
    FROM generate_series($3::timestamptz, $4::timestamptz, $1::interval) AS g(bucket_time)
)
```

- `server/src/internal/metrics/query.go`: both `all_buckets` CTEs.

- `server/src/internal/metrics/query_test.go`: the existing derived
  query assertion checked only the `generate_series(...)` substring,
  which both forms contain; it now asserts the full `FROM
  generate_series(...) AS g(bucket_time)` form. An equivalent
  assertion is added to `TestBuildMetricsQuery` so the placement is
  pinned at both sites.

## Testing

- `cd server && make test` passes; `gofmt`, `go vet`, and
  `golangci-lint run ./internal/metrics/...` are all clean.

- `BuildMetricsQuery` and `BuildDerivedMetricsQuery` are both at 100%
  statement coverage in `go tool cover -func`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metrics query compatibility while preserving gap-filled time buckets for standard and derived metrics.
* **Tests**
  * Updated coverage to verify the generated time-series buckets are constructed correctly.

### DIFF
--- a/server/src/internal/metrics/query.go
+++ b/server/src/internal/metrics/query.go
@@ -427,7 +427,8 @@ func BuildMetricsQuery(
             GROUP BY date_bin($1::interval, collected_at, $3)
         ),
         all_buckets AS (
-            SELECT generate_series($3::timestamptz, $4::timestamptz, $1::interval) AS bucket_time
+            SELECT bucket_time
+            FROM generate_series($3::timestamptz, $4::timestamptz, $1::interval) AS g(bucket_time)
         )
         SELECT
             all_buckets.bucket_time,
@@ -682,7 +683,8 @@ func BuildDerivedMetricsQuery(
 
 	ctes = append(ctes, `
         all_buckets AS (
-            SELECT generate_series($3::timestamptz, $4::timestamptz, $1::interval) AS bucket_time
+            SELECT bucket_time
+            FROM generate_series($3::timestamptz, $4::timestamptz, $1::interval) AS g(bucket_time)
         )`)
 
 	query := fmt.Sprintf(`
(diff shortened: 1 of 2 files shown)


AntTheLimey (approved):
**Optional:** neither `all_buckets` CTE (query.go around line 429 and
line 685) has a comment saying why `generate_series` is in `FROM`
rather than the select list. The reason is DuckDB-specific
compatibility (a set-returning function in select-list position
returns a LIST there, not a set) and is not derivable from the code;
it currently only lives in the PR description. Worth a one-line
comment at each site so a later "simplification" back to the
select-list form doesn't silently reintroduce the DuckDB break.

Everything else checks out: verified the two forms are row-for-row
identical in Postgres across start==end, start>end, and a
non-dividing interval (partial trailing bucket excluded in both);
param numbering and the `bucket_time` alias are unchanged; both
`BuildMetricsQuery` and `BuildDerivedMetricsQuery` are at 100%
coverage under the DB-backed test run; `gofmt` and `make lint` are
clean; no other select-list `generate_series` site was left behind.

Issue #441, opened 2026-09-10
Collector runs all past-due probes at once on startup, with no stagger or concurrency cap


dpage:
On startup the collector runs every past-due probe immediately, for every
monitored connection, with no stagger, jitter or concurrency cap. On a
deployment of any size that is a thundering herd large enough to exhaust
a modest memory limit, and because an OOMKill restarts the process and
puts every probe back in the past-due state, it self-sustains.

Reported as part of #437, where it combined with the garbage collector's
uptime-coupled schedule to fill a datastore disk: the restart loop meant
the dropper's five-minute startup delay never elapsed, so no partition was
ever dropped. Neither defect alone fills a disk; together they guarantee
it.

## Cause

`scheduleProbeForConnection` in `collector/src/scheduler/scheduler.go`
calls `calculateInitialDelay` and then runs the probe immediately whenever
the delay is not positive. `calculateInitialDelay` returns 0 when there is
no recorded last collection time (and also on either error path), and a
negative delay for any probe already past due.

After a restart, every probe whose interval is shorter than the outage is
past due, which for 30-second probes is all of them. Each runs in its own
goroutine with no admission control, so concurrency scales with
connections multiplied by probes.

In the environment reported in #437 that is 5 connections at 20-30 probes
each, so roughly 100-150 probe executions starting at once, against a
256Mi limit.

## Fix

Spread the first execution rather than firing everything at once. A
per-probe jitter across the probe's own interval would be the least
surprising option, since it also de-clusters steady-state collection, and
it should be combined with a cap on how many probes may execute
concurrently so that the worst case does not scale with the number of
monitored connections.

Worth checking the memory profile of a startup burst separately: 256Mi is
not an unreasonable limit for this workload, so if 100 concurrent probes
genuinely need more than that, the per-probe footprint deserves a look
too.

Issue #62, opened 2026-04-20
Partition dropper fails with "conn busy" when competing with metric probes for connections


imtiazqa:
The collector logs warnings when trying to drop expired partitions while metric probes are actively using the connection pool:

```
2026/04/20 10:02:11 Collector is running. Press Ctrl+C to stop.
2026/04/20 10:07:11 Warning: failed to drop partition pg_connectivity_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_database_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_activity_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_all_indexes_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_all_tables_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_checkpointer_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_connection_security_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_database_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_database_conflicts_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_io_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_recovery_prefetch_20260406: conn busy
2026/04/20 10:07:11 Warning: failed to drop partition pg_stat_wal_20260406: conn busy

```
The partitions are eventually dropped after a collector restart when connections are free, confirming this is a connection pool contention issue — the partition dropper competes with probes for the same connections.

**Expected Behavior:**

The partition dropper should use a dedicated connection or retry with backoff rather than failing silently when the pool is busy.

imtiazqa:
create a new test partition to verify with this current run:


sudo -u postgres psql -d ai_workbench -c "
CREATE TABLE IF NOT EXISTS metrics.pg_stat_activity_20260413 
PARTITION OF metrics.pg_stat_activity
FOR VALUES FROM ('2026-04-13') TO ('2026-04-14');"
Then wait 5 minutes and check:


sudo -u postgres psql -d ai_workbench -c "
SELECT tablename FROM pg_tables 
WHERE schemaname = 'metrics' AND tablename LIKE '%20260413%';"

[...]

PR #442, opened 2026-09-10
Time partition retention from its last run, not from process start

### LINKED ISSUE #437: Partition dropper does not enforce retention_days — partitions persist for weeks past configured window, filling disk
### Summary

  The collector's partition dropper is not enforcing `probe_configs.retention_days`. Partitions
  remain long past their configured retention window despite hundreds of collector restarts,
  and in our case this filled the metrics datastore's disk and crash-looped it.

  This looks related to #62 ("Partition dropper fails with 'conn busy' when competing with
  metric probes for connections"), which was closed as fixed on 2026-04-20 with the note that
  partitions are "eventually dropped after a collector restart when connections are free." Our
  evidence below suggests that either the fix doesn't hold under a smaller connection pool, or
  this is a distinct/regressed issue — the partitions here never get dropped, even after 100+
  restarts.

  ### Environment
  
  - `ai-dba-collector` image: `ghcr.io/pgedge/ai-dba-collector:latest`, reports `v1.0.0` at startup                                                                   
  - `pool.datastore_max_connections: 10`, `pool.max_connections_per_server: 3`
  - 5 monitored connections, ~20-30 probes each, intervals ranging 30s–3600s

  ### Expected behavior

  A partition should be dropped once it ages past its probe's `probe_configs.retention_days`.                                      
                                                                                                                                                                      
  ### Actual behavior

  `metrics.pg_stat_statements_20260831` — a weekly partition covering 2026-08-31 to 2026-09-07 —
  was still present and growing on 2026-09-09 (9+ days after its retention window closed), at
  5,194,085 rows / 3.9GB. `pg_stat_statements`'s configured `retention_days` is 7
  (`probe_configs.retention_days`).

  In the interim, `ai-dba-collector` restarted 112+ times (many due to a separate OOMKill issue
  we also hit — startup fires all probes for all 5 connections at once with no stagger, spiking
  past the 256Mi memory limit). Per #62's own description, restarts should free connections and
  let the dropper catch up — that did not happen here across 100+ restarts.

  Net effect: the `ai-dba-postgres` datastore PVC (10Gi) filled to 100%, and `ai-dba-postgres`
  crash-looped continuously (`FATAL: could not write to file "pg_wal/xlogtemp.N": No space left
  on device` on every WAL-recovery attempt at startup), taking the whole Workbench down.

  ### Other stale partitions observed at the same time (same symptom, various probes)

  pg_stat_activity_20260831, pg_connectivity_20260831, pg_stat_replication_20260831,
  pg_database_20260831, pg_stat_database_20260831, pg_statio_all_sequences_20260831,
  pg_stat_connection_security_20260831, pg_stat_database_conflicts_20260831,
  pg_stat_all_tables_20260831, pg_stat_all_indexes_20260831, pg_stat_wal_20260831,
  pg_stat_checkpointer_20260831, pg_stat_io_20260831, pg_stat_recovery_prefetch_20260831
  — all still present past their 7-day `retention_days` as of 2026-09-09.

  ### Additional evidence

  Checked the collector's current and previous container logs (kubectl logs / --previous) for any partition-dropper activity. No mention of connbusy, drop partition, or similar, in either.
  Given each collector pod only survives a few minutes before restarting (OOM or otherwise), this doesn't rule out sporadic attempts, but there's no log evidence the dropper is running or failing at all, as opposed to #62 where failures were at least logged as warnings.

  ### Workaround in place
  
  We deployed an external CronJob that drops any `metrics.*_<YYYYMMDD>` partition past its                                                                            
  probe's `retention_days` (read from `probe_configs`), and separately raised the collector's
  memory limit to 512Mi to stop the OOM restarts. This masks the symptom but doesn't fix the
  underlying dropper.

  ### Suggested next step
                                                                                                                                                                      
  Re-open #62, or treat this as confirmation the earlier fix is incomplete — worth checking
  whether the dropper's retry/backoff still starves under a small connection pool with many
  concurrent probes, or whether it's silently erroring/exiting rather than retrying at all.

dpage:
Fixes #437.

The partition dropper never enforced `retention_days` for the reporter,
and the reason is that it never ran. `GarbageCollector.run` waited five
minutes after process start before its first pass, then ticked every 24
hours, and kept no record of when it had last run. Retention was
therefore a function of unbroken process uptime: a collector dying inside
five minutes collected garbage zero times, and each restart put it back
to the beginning of the delay. With pods OOM-restarting every few minutes
across 112+ restarts, the first pass genuinely never happened, which is
why the reporter found no dropper activity in the logs at all, and why
the datastore's disk filled and crash-looped.

Worth being clear about two things this is not. The retention logic was
never at fault: given `retention_days = 7` and two partitions, one 30
days old and one current, it drops precisely the old one. And this is not
a regression of #62; that fix is present and correct in the released
code, and no `conn busy` is involved.

## What changed

Completions are recorded in a new `maintenance_runs` table (migration 8)
and the next pass is timed from that record rather than from process
start, so a restart resumes the existing cycle instead of beginning a new
one. A task that has never run, or whose record cannot be read, counts as
due: collecting slightly early is harmless, whereas not collecting is
what fills a disk. The startup grace period drops from five minutes to
30 seconds, which is enough to stay clear of the probe burst the
scheduler runs at startup without reintroducing the original bug, and a
failed pass backs off rather than retrying immediately so that an
unreachable datastore cannot produce a tight error loop.

This also closes the same hole for rolling deployments and node drains,
which fail identically today and are considerably more common than an
OOM loop.

The cadence moved onto struct fields seeded from the constants so the
scheduling loop can be driven in tests without waiting out real delays.

## Notes for review

- The v7 migration tests rewound only their own `schema_version` row,
  which stops working as soon as a later migration exists, because
  `getCurrentVersion` takes `MAX(version)` and v7 was consequently
  skipped rather than re-run. They now rewind from v7 onwards, which is
  robust to future migrations.
- `MinCollectorSchemaVersion` in the server is 4, so migration 8 needs no
  compatibility change there.
- Two things noticed in passing and deliberately left alone: a probe
  config naming a metrics table that does not exist is silently a no-op
  rather than an error, since the partition lookups simply match no rows;
  and the startup thundering herd that caused the reporter's OOM loop is
  filed separately as #441, because it is a distinct defect that happens
  to have combined with this one.

## Testing

`collector/make test` passes in full, including lint. Coverage on the
changed files is 95.7% for `garbage_collector.go` and 100% for
`maintenance.go`, both above the 90% floor.

New tests cover the scheduling decision directly (never run, unreadable
record, overdue, mid-cycle, remainder inside the grace period, a run
recorded in the future, and a failed attempt backing off), a 112-restart
loop asserting the delay never depends on process start, an end-to-end
pass against a real datastore that records its completion and is then
seen mid-cycle by a fresh collector instance, a full pass that drops a
partition past its configured `retention_days`, and the failure branches
for an unreachable datastore, unloadable probe configs, a probe with no
metrics table, and an unrecordable completion.

`make test-all` does not currently pass on `main`:
`TestStoreMemoryGeneratesEmbeddingIntegration` and
`TestRecallMemoriesGeneratesQueryEmbeddingIntegration` fail with
`expected 3 dimensions, not 4000` on any host that has pgvector
installed, which open PR #382 fixes. I confirmed both fail identically on
a pristine `origin/main` checkout. Separately, running the server's
packages in one parallel `go test` invocation fails a batch of
`TestClusterHandler_*` integration tests through cross-package
interference on the shared database; that also reproduces on pristine
`main` and passes when run per-package as the Makefile does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Partition retention cleanup now persists successful maintenance runs and schedules future runs based on the last completed run rather than process uptime.
  * Overdue or unrecorded cleanup runs execute after startup, including after restarts and rolling deployments.
  * Failed maintenance runs retry with a backoff delay.
  * Shutdown and cancellation are handled safely, while missing or unusable datastore configurations return errors instead of causing crashes.

* **Documentation**
  * Updated guidance and changelog information for partition retention scheduling and recovery behavior.

### DIFF
--- a/.claude/golang-expert/partitioning.md
+++ b/.claude/golang-expert/partitioning.md
@@ -133,3 +133,37 @@ Rules for the datastore connection shared across a GC pass:
   client-side protocol violation, not transient contention, and
   retries will only succeed because the loop body has moved past
   the busy state — the underlying bug stays.
+
+## Scheduling Retention: Never Couple It to Process Uptime
+
+Retention enforcement must not be scheduled from process start.
+The garbage collector originally waited five minutes after startup
+and then ticked every 24 hours, holding no record of when it last
+ran. That made retention a function of *unbroken uptime*: a
+collector restarting more often than the startup delay never
+collected garbage at all, and every restart put it back to the
+beginning of the delay. In issue #437 an OOM restart loop therefore
+filled a datastore's disk, because partitions were never dropped
+across more than a hundred restarts.
+
+The pattern to follow instead:
+
+- Persist each completed pass. `maintenance_runs` holds one row per
+  task (`database.PartitionRetentionTask` for the dropper), written
+  by `database.RecordMaintenanceRun`.
+- Time the next pass from that record, not from process start, so a
+  restart resumes the cycle rather than beginning a new one. See
+  `computeNextRunDelay` in `collector/src/garbage_collector.go`.
+- Treat "never run" and "cannot read the record" as due now. Failing
+  towards collecting slightly early is harmless; failing towards not
+  collecting is what fills a disk.
+- Keep the startup grace period short, and only long enough to stay
+  clear of the probe burst the scheduler runs at startup. A long
+  grace period reintroduces the original bug for any process that
+  does not live that long.
+- Back a failed pass off rather than retrying immediately, so an
+  unreachable datastore does not produce a tight error loop.
+
+The same reasoning applies to any periodic maintenance added later:
+if the schedule lives only in memory, an unstable process silently
+stops doing the work, and nothing in the logs says so.
--- a/collector/src/database/maintenance.go
+++ b/collector/src/database/maintenance.go
@@ -1,0 +1,58 @@
+/*-------------------------------------------------------------------------
+ *
+ * pgEdge AI DBA Workbench
+ *
+ * Copyright (c) 2025 - 2026, pgEdge, Inc.
+ * This software is released under The PostgreSQL License
+ *
+ *-------------------------------------------------------------------------
+ */
+package database
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+// PartitionRetentionTask is the maintenance_runs key under which the
+// partition dropper records its completions.
+const PartitionRetentionTask = "partition_retention"
+
+// GetMaintenanceRun returns when the named maintenance task last
+// completed. The second return value is false when the task has never
+// run, which is distinct from an error and is the expected state on a
+// freshly migrated datastore.
+func GetMaintenanceRun(ctx context.Context, conn *pgxpool.Conn, task string) (time.Time, bool, error) {
+	var lastRun time.Time
+	err := conn.QueryRow(ctx, `
+        SELECT last_run_at FROM maintenance_runs WHERE task = $1
+    `, task).Scan(&lastRun)
+
+	if errors.Is(err, pgx.ErrNoRows) {
+		return time.Time{}, false, nil
+	}
+	if err != nil {
+		return time.Time{}, false, err
+	}
+
+	return lastRun, true, nil
+}
+
+// RecordMaintenanceRun stamps the named maintenance task as having
+// completed at the given time. Storing the completion in the datastore
+// rather than in memory is what decouples the schedule from the
+// collector's process uptime, so a restart resumes the existing cycle
+// instead of starting a fresh one.
+func RecordMaintenanceRun(ctx context.Context, conn *pgxpool.Conn, task string, at time.Time) error {
+	_, err := conn.Exec(ctx, `
+        INSERT INTO maintenance_runs (task, last_run_at)
+        VALUES ($1, $2)
+        ON CONFLICT (task) DO UPDATE SET last_run_at = EXCLUDED.last_run_at
+    `, task, at.UTC())
+
+	return err
+}
--- a/collector/src/database/maintenance_test.go
+++ b/collector/src/database/maintenance_test.go
@@ -1,0 +1,150 @@
+/*-------------------------------------------------------------------------
+ *
+ * pgEdge AI DBA Workbench
+ *
+ * Copyright (c) 2025 - 2026, pgEdge, Inc.
+ * This software is released under The PostgreSQL License
+ *
+ *-------------------------------------------------------------------------
+ */
+package database
+
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+// TestMaintenanceRunRoundTrip covers the persistence that decouples the
+// partition dropper's schedule from process uptime (issue #437).
+func TestMaintenanceRunRoundTrip(t *testing.T) {
+	if err := setupTestDatabase(); err != nil {
+		t.Skipf("test database not provisioned: %v", err)
+	}
+
+	pool, conn := getTestConnection(t)
+	defer pool.Close()
+	defer conn.Release()
+
+	ctx := context.Background()
+
+	sm := NewSchemaManager()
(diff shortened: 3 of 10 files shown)


dpage:
Renumbered the `maintenance_runs` migration from 8 to 9, since #418 claimed 8 first for its alert-rule semantics changes.

One merge-order constraint that follows: **this needs to merge after #418.** `getCurrentVersion` takes `MAX(version)`, so a datastore that applied 9 while 8 did not yet exist would skip 8 permanently once it appeared. Merging in that order means `main` never carries the gap.

Also worth noting that #418 independently made the same v7 migration-test fix this PR carries (rewinding `schema_version` from v7 onwards rather than just the v7 row, since `MAX(version)` otherwise skips the re-run). Whichever lands second will see that hunk as an identical-content conflict.

AntTheLimey (approved):
No blocking issues. I checked the retention-window edge cases in the
brief directly against the code: `DropExpiredPartitions` computes its
cutoff from `time.Now()` at drop time rather than from the recorded
last-run timestamp, so a stale, missing, or unreadable
`maintenance_runs` row only changes how soon the next pass is
attempted, never what gets dropped — there's no path here to dropping
a partition still inside its retention window, across a restart or
across two collectors racing into `collectGarbage` at once (the
upsert in `RecordMaintenanceRun` has no read-modify-write window, and
`DROP TABLE IF EXISTS` makes a duplicate drop attempt harmless). I
also traced the migration-numbering question CodeRabbit raised:
version 8 already belongs to #418 (merged, and an ancestor of this
head), this PR's migration correctly renumbers to 9, `LatestVersion()`
comes out to 9, and `migration_v8_test.go` still correctly exercises
migration 8 — that file's own diff is just relaxing an assertion that
could never hold once a migration 9 exists. Migration 9 follows the
repo's idempotency and `COMMENT ON` conventions and the full suite
re-runs it twice via `TestMigrateIdempotency` without issue. Two
non-blocking notes. CLAUDE.md asks for the ratchet to be raised whenever
a change lifts the total, and this branch measures 87.7% against 86.0%
on `origin/main`, so `collector/Makefile`'s `COVERAGE_MIN` is a
candidate — but I would not raise it to 87.7 yet. That figure is from a
single local run, while the ratchet is enforced per matrix entry, and
the collector's PG 14 and PG 15 jobs currently report around 85.4%
because version-gated tests skip there. Raising the floor to 87.7 would
turn two already-red jobs into permanently red ones. The matrix minimum
is the number that matters, and it is worth settling separately from
this PR. Second note: an untouched test's doc comment still cites the
old five-minute startup delay instead of the new 30-second grace.
gofmt, lint and `make coverage` all pass on the branch. Textually this
only collides with #420, #419 and #386 in `docs/changelog.md`; I
merge-tree'd all three and built the merged `collector` tree in each
case with no compile errors, and #420's own flagged
`tx.Rollback(context.Background())` pattern does not appear anywhere
in this PR's new code. This PR's only real dependency, #418, is
already on `main`, so there's no merge-order constraint left to
enforce.

dpage:
Rebased onto `main` and pushed 82dc8416 to clear the Codacy gate: Opengrep flagged the test database name being concatenated into the `CREATE DATABASE` and `DROP DATABASE` calls in `garbage_collector_integration_test.go`. The name is now quoted with `pgx.Identifier{name}.Sanitize()` and the two statements are assembled into variables before the `Exec` calls; the other five findings in that file are on constant literals and should fall away with the concatenation gone. Test behaviour is unchanged.

dpage:
Rebased again after #417 landed (changelog only). Codacy: with your agreement I marked the seven Opengrep SQL-injection findings on `collector/src/garbage_collector_integration_test.go` as false positives in Codacy. Two are the test database `CREATE`/`DROP` of a timestamp-derived name, now quoted with `pgx.Identifier.Sanitize()` since an identifier cannot be bound as a parameter; the other five are constant string literals with no concatenation or external input at all.

PR #451, opened 2026-09-14
Centralise LLM chat request construction in the client

### LINKED ISSUE #373: Centralise /api/v1/llm/chat request construction to enforce the wire contract
## Background

PR #372 fixed #370, where every chat request carrying a tool list was rejected by Anthropic because the client forwarded its internal camelCase `inputSchema` verbatim to `POST /api/v1/llm/chat`, whose body is decoded into the vendored `pgedge-go-llm-lib` `llm.Tool` struct tagged `json:"input_schema"` (snake_case). The mismatched key silently unmarshalled to an empty schema for every tool.

The immediate fix added a `normaliseTools` helper and applied it at the two call sites that forward tools.

## The underlying fragility

The root cause is not the casing itself but that the chat request body is hand-built independently at multiple sites, each separately responsible for remembering to normalise both `messages` and `tools` to the library's wire contract. There are currently four direct callers of `/api/v1/llm/chat`:

- `client/src/hooks/chat/chatAgenticLoop.ts` (forwards tools; fixed in #372)
- `client/src/utils/agenticLoop.ts` (forwards tools; fixed in #372)
- `client/src/hooks/useChartAnalysis.ts` (no tools today)
- `client/src/hooks/useQueryOverview.ts` (no tools today)

This class of bug arose precisely because the earlier wire-contract migration added `normaliseMessages` but missed `tools` at the hand-rolled sites. The same omission can recur the next time someone adds a `tools` list to a new call site, or adds a new field to the wire contract.

## Proposed follow-up

Introduce a single shared request-builder (or thin client wrapper) for `/api/v1/llm/chat` that always normalises `messages` and `tools` to the library's wire shape, and route all four call sites through it. That makes the wire contract impossible to get wrong per-site and removes the duplicated body-construction logic.

Non-urgent; noted during review of #372.

dpage:
## Summary

Every caller of `POST /api/v1/llm/chat` hand-built its own request body and had to remember to normalise both `messages` and `tools` to the `pgedge-go-llm-lib` wire contract, which is how #370 slipped through. This adds `client/src/utils/llmChat.ts` with `LLM_CHAT_PATH` and `buildChatRequestInit()` as the single place that knows the wire shape, and routes the chat panel, the analysis agentic loop, chart analysis and query overview through it.

The only wire-level difference is that the chat panel now omits `tools` when the list is empty rather than sending `[]`; the library treats nil and empty tools identically.

## Test plan

- [x] New unit tests for the builder in `client/src/utils/__tests__/llmChat.test.ts`
- [x] Existing tests for the four call sites pass unchanged
- [x] `npm run lint` clean

Closes #373

## Summary by CodeRabbit

- **Improvements**
  - Standardized chat, chart analysis, and query overview requests for more consistent communication with the LLM service.
  - Improved handling of message content, system prompts, tools, and request cancellation across supported experiences.
  - Empty tool configurations are now omitted from requests, helping keep request payloads concise.

- **Documentation**
  - Added guidance describing the required format and handling for LLM chat requests.

- **Tests**
  - Expanded automated coverage for request formatting, tool conversion, system prompts, and cancellation behavior.


dpage:
@coderabbitai review

AntTheLimey (approved):
This centralises the `/api/v1/llm/chat` request shape well. I checked
the specific claim in the description against the server and
`pgedge-go-llm-lib`: every provider (`anthropic`, `gemini`, `openai`,
`ollama`) gates tool inclusion on `len(req.Tools) > 0`, never on a
nil check, so a nil and an empty `tools` slice really are
indistinguishable on the wire — the chat panel's new omit-when-empty
behaviour is safe. All four call sites now go through
`buildChatRequestInit`, no hand-built body remains, and the one call
site that ever supported request cancellation
(`chatAgenticLoop.ts`) still forwards its `AbortSignal` through the
builder; the other three never had cancellation and still don't.
`llmChat.ts` and its four callers land at 90%+ line coverage each
(scoped Vitest run; the full `make coverage` hit the known
`ServerDialog`/`AdminTokenScopes` 5000ms flake, unrelated to this
PR), lint is clean on the touched files, and the typecheck error
count is unchanged at 575 between this branch's merge-base and its
head. The `quality-checklist.md` addition is scoped to what this
change actually does. No merge-order concern against #457 — it edits
a different section of the same checklist file and touches none of
the other six files here.

PR #455, opened 2026-09-14
Count deadlocks and temp files per hour in the alerter, enforce required_extension, retire table_bloat_ratio

### LINKED ISSUE #409: Alerter reimplements the server's rate derivation, and its deltas are probe-interval dependent

dpage:
## Summary

The alerter's `deadlocks_delta` and `temp_files_delta` metrics reported the change between two consecutive samples, so the `deadlocks_detected` and `temp_files_created` thresholds silently scaled with the probe interval. Both now sum positive per-sample deltas over a fixed one hour window, as `checkpoint_warning` and `wal_archive_failed` already do since #406, and their historical variants bucket by hour so baselines measure the same quantity. Migration 11 updates the units to `deadlocks/hour` and `files/hour` and leaves thresholds untouched. With every remaining delta on a fixed window no rule is per-probe-interval any more.

`required_extension` was stored on rules but never enforced. Rules that carry one are now evaluated only on connections whose newest `metrics.pg_extension` snapshot lists the extension, and an alert on a connection lacking it is left active rather than resolved, so "extension missing" is no longer indistinguishable from "condition resolved".

`table_bloat_ratio` duplicated `dead_tuple_ratio` with a different denominator; its registry entry is removed, the built-in rule is disabled by the migration and its open alerts are cleared. The rule row stays so historical alerts remain attributable.

The shared rate-derivation package discussed on the issue is deliberately not attempted here; see the proposal comment on #409 for the reasoning.

The review follow-up carries the raw sample count through `HistoricalMetricValue` so baseline warmup still counts samples rather than hourly buckets, moves the historical filter inside the aggregate so a single-sample connection keeps its zero bucket, cuts the buckets on the UTC value of `collected_at`, resolves the `required_extension` gate once per rule in `cleanResolvedAlerts`, and states the operator-visible sensitivity change to `temp_files_created` in the changelog and rule reference.

## Test plan

- [x] `make test-all` in `pkg`, `collector`, `server` and `alerter` against local Postgres, lint clean (alerter total 71.0%, ratchet raised from 69 to 70; collector total 88.0%, ratchet raised from 86 to 87)
- [x] New integration tests: interval independence, counter reset, single sample, hourly buckets, newest-snapshot extension lookup, evaluation and resolution gates, migration 11 upgrade path

Closes #409

## Summary by CodeRabbit

- **New Features**
  - Alert rules can require a PostgreSQL extension; alerts apply only to connections whose latest snapshot includes it.
  - Deadlock and temporary-file metrics now report hourly counts using positive counter changes, independent of collection frequency.

- **Changes**
  - The `table_bloat_ratio` rule is retired, disabled by default, and no longer collected. Existing active alerts are cleared during upgrade.

- **Documentation**
  - Updated alert-rule references and migration guidance explain extension requirements, hourly metric behavior, thresholds, and the retired table-bloat rule.


dpage:
@coderabbitai review

AntTheLimey (changes requested):
Three things to fix. One is a documentation gap that bites hard on upgrade. The
other two are both in the new `historicalSQL`.

**The `temp_files_created` rule gets about twelve times more sensitive when
this ships, and the changelog says it doesn't.** The threshold stays at 100.
What it is compared against changes, from "the biggest jump between two
samples" to "the total for the whole hour". The shipped `pg_stat_database`
interval is 300 s, so twelve samples now land in the window.

Concretely: a database steadily making 300 temp files an hour used to report
about 25 and never fire. It now reports 300 and fires every cycle. Operators
who tuned the threshold get the same shift, which
`TestMigrationV10_UpgradesLegacyRowsAndClearsRetiredAlerts` pins.

I'm not arguing with the threshold. Your #409 proposal makes the case for 100
an hour and I accept it. The problem is that the changelog says "The default
thresholds of 0 deadlocks and 100 temporary files are unchanged, as are any
operator overrides". That is true as written, but an operator reads it as
"nothing changes for me". Please say in the changelog and in the High Temporary
File Usage section of the rule reference that the compared value is now an
hourly total, that the rule will fire on databases where it previously didn't,
and that anyone wanting the old sensitivity should multiply their threshold by
3600 divided by the probe interval. `deadlocks_detected` at `> 0` is genuinely
unaffected, since any positive delta fired under both.

**Making `historicalSQL` hourly stops the anomaly baselines warming up.**
`GetHistoricalMetricValues` now returns one row per hour instead of one per
sample, `calculateAllBaseline` stores that row count as `SampleCount`, and
`isBaselineWarm` checks it against `anomaly.tier1.warmup.all.min_samples`,
which defaults to 100.

At the default `lookback_days` of 7 that count drops from roughly 2,016 to at
most 168. So these two metrics need about 100 hours of unbroken collection to
warm, not the 24 hours `min_span_hours` implies. At `lookback_days` of 4 or
less, which is legal and documented, the ceiling is 96 and the baseline can
never warm at all. The hourly and daily baseline rows can't cover for it,
because `GetMetricBaselines` orders by `period_type` and the detector reads the
`all` row.

Two ways out: carry the underlying sample count through `HistoricalMetricValue`
so `SampleCount` still means what the warmup config assumes, or document the
new relationship between `lookback_days` and `min_samples` so an operator can
see why detection went quiet.

**`historicalSQL` also drops a bucket that `latestSQL` would report as zero.**
`latestSQL` filters inside the aggregate, with `COALESCE(SUM( GREATEST(x -
prev_x, 0)) FILTER (WHERE prev_x IS NOT NULL), 0)`, so a connection whose only
sample has no predecessor reports 0. `historicalSQL` puts `WHERE prev_x IS NOT
NULL` before the `GROUP BY` instead, so that row is removed and its bucket
produces nothing.

This one is narrow and I want to be straight about that. `LAG` runs across the
whole lookback, so only one row per connection and database has a null
predecessor, and a bucket only disappears when the leftmost one holds a single
sample. The case that reliably bites is a newly onboarded connection with one
sample, which gets no historical row at all. It's worth fixing because the
comment above the entry promises both "a connection with a single sample
reports 0 instead of vanishing" and that the historical query feeds baselines
the same per-hour figure. Moving the filter inside the aggregate, as
`latestSQL` already does, covers it.

Three smaller things while you're in there.

The hour buckets use `date_trunc('hour', collected_at)`, which truncates in the
session's `TimeZone`. Nothing in the alerter's pool config pins that to UTC,
and the collector writes UTC timestamps, so on a deployment whose `timezone`
GUC isn't UTC the boundaries shift against the stored data. `date_trunc('hour',
collected_at AT TIME ZONE 'UTC')` makes it deployment-independent.

The `metrics-queries.md` issue history gains two #409 bullets that say the same
thing; the second is wholly contained in the first.

`alertRequiredExtensionMissing` calls `GetAlertRuleByID` for every active
threshold alert on every cleanup pass, before it knows whether the rule is
gated at all, then resolves the extension set per alert. The evaluator already
does this once per rule. Hoisting the lookup into `cleanResolvedAlerts`, which
already holds the alert list, would match it.

Everything else holds up, and I checked it rather than assuming. The hourly
window is built correctly: `LAG` in the inner CTE partitioned by connection and
database, `SUM` outside it, so you difference first and sum second rather than
differencing an aggregate. The latest query matches the archiver and
checkpointer entries statement for statement, and the new integration test pins
counter reset, single sample, a previous sample outside the window,
per-database separation and the hour-boundary bucket. The `required_extension`
gate does what the description says on both paths: the resolution check returns
before `GetLatestMetricValues`, so a missing extension can't reach the branch
that clears on absent data, both lookup failures fall through ungated, and
neither pointer dereference is reachable with nil. Migration 10 is safe when
the rule is absent, touches no partitioned table, and `default_enabled = FALSE`
really does disable the rule since `GetEnabledAlertRules` filters on that
column alone; no `table_bloat_ratio` reference survives elsewhere. On the SQL
itself, nothing casts or wraps an indexed column in a predicate, so the
`queryid::text` problem doesn't recur, and every runtime value binds as a
placeholder. Measured: alerter coverage 69.8% against the 69 ratchet, up from
69.4% on main, new functions at 92.3-100%; collector `make test-all` passes at
86.7% against its 86 ratchet; both sub-projects gofmt clean, zero lint issues;
19 new tests pass against PostgreSQL 17.10.

On merge order, #457 also registers collector migration 10, so the two collide:
`git merge-tree` gives an add/add conflict on `migration_v10_test.go` and
content conflicts in `schema.go` and `metrics-queries.md`. The migrations are
semantically independent, so only the number is in dispute.

I'd merge #457 first and renumber this one to 11, because a missed version
literal fails loudly here and quietly there. This PR's `migrationV10` helper
calls `t.Fatal` when 10 isn't registered. #457's
`TestMigrationV10_ColumnCascadesToExistingPartition` and
`TestMigrationV10_Idempotent` both rewind with `DELETE FROM schema_version
WHERE version >= 10`, and the second then asserts `COUNT(*) ... WHERE version =
10` is 1. So if #457 renumbers to 11 and those literals stay at 10, the rewind
deletes this PR's row and the assertion measures the wrong migration. The test
keeps passing while testing nothing.

One thing to watch whichever order you take: `getCurrentVersion` returns
`MAX(version)` and migrations are forward-only, so any datastore that already
ran the loser's migration 10 will permanently skip the winner's. That means a
reused CI database, the demo stack, or a preview environment. It'd sit there
with the old units, descriptions and `table_bloat_ratio` state while the
alerter already evaluates hourly sums, and nothing detects it. Worth recreating
those rather than trusting the version counter.

The duplication between this PR and #457 isn't a finding, for the record. Your
proposal comment on #409 makes the case for deferring the shared package and I
agree with it.

dpage:
All three blocking findings confirmed and fixed, plus the three smaller ones
and the renumber. Rebased onto `main` and force-pushed.

**Sensitivity wording.** Your arithmetic checks out: the shipped interval is
300s (`collector/src/database/schema.go:497`) against a seeded threshold of
100, so twelve samples land in the hour and the rule really does get about
twelve times more sensitive. The changelog said the true thing and left the
wrong impression, which is the worse failure of the two.
`docs/changelog.md:132-147` now states that the compared value is an hourly
total, that `temp_files_created` will fire on databases where it previously did
not, gives your 300-files-an-hour example, and gives the restore rule (multiply
by 3600 ÷ sampling interval, so 1200 at the default). The High Temporary File
Usage section of the rule reference carries the same, and I have added an
explicit line saying Deadlocks Detected is unaffected, since a threshold of 0
fired on any positive change under both comparisons.

**Baseline warmup.** Confirmed the whole chain, including that
`GetMetricBaselines` orders by `period_type` so the hourly and daily rows
cannot stand in. Took the first of your two options, since documenting a
`lookback_days` of 4 as silently disabling detection is not a fix.
`HistoricalMetricValue` now carries `SampleCount`
(`alerter/src/internal/database/types.go:181-197`), the bucketed queries return
`COUNT(*)` per bucket, and `totalSamples` (`engine/baselines.go:319`) sums it,
treating an unset or negative count as one sample so every other metric and
every hand-built value behaves exactly as before. I applied it to the hourly
and daily builders too, so `sample_count` means the same thing in every row;
their thresholds of 5 and 3 were already met by bucket counts, so no gating
behaviour moves there.

**Dropped zero bucket.** Fixed as you suggested: both historical queries now
use `SUM(GREATEST(x - prev_x, 0)) FILTER (WHERE prev_x IS NOT NULL)` with no
pre-`GROUP BY` `WHERE`, matching `latestSQL` statement for statement. Pinned by
`TestMetricRegistry_HourlyDeltas_HistoricalKeepsZeroBucket`. You were right to
flag it as narrow, but the newly onboarded connection getting no historical row
at all is the case that would have been genuinely confusing to debug.

**Session TimeZone.** Buckets are now cut as `date_trunc('hour', collected_at
AT TIME ZONE 'UTC') AT TIME ZONE 'UTC'`, which keeps the result `timestamptz`
so the scan type and the hour-of-day grouping downstream are unchanged. The
test pins the session TimeZone through `RuntimeParams` using Asia/Kolkata,
whose half-hour offset splits a UTC hour and so catches the bug that a
whole-hour offset would hide; reverting to plain `date_trunc` makes it fail
with two buckets at 06:30 and 07:30 rather than one.

**The other two.** Redundant `metrics-queries.md` bullet removed.
`alertRequiredExtensionMissing` is gone, replaced by `resolveExtensionGates`
(`engine/cleanup.go:58`), which loads each distinct rule once and each distinct
extension once per pass and passes the connection set into
`checkAlertResolved`, matching what the evaluator already did. A lookup failure
still leaves the rule ungated.

**Renumbered to 11**, with `migration_v10_test.go` renamed and every literal and
helper moved with it, including the `t.Fatal` on an unregistered version.
`schema_test.go` expects `{1..9, 11}` with a comment explaining the reserved
slot, and `Migrate()` tolerates the gap. Once #457 lands the expectation closes
back up to `{1..11}`.

One thing you did not raise but which belongs in the record: while getting the
engine suite to run against a per-worktree database I first relaxed
`assertLocalTestDSN` to a `strings.HasPrefix` check, which is the exact defect
that got #467 closed this morning, since `ai_workbench_prod` is the example
production database name in five places on `main`. It now uses the anchored
regex from #462 with that PR's refusal cases mirrored, so the two reconcile
when #462 lands its shared `requireLocalTestDSN`.

**On the version counter.** Your warning is the right one and I have no clever
answer to it: `getCurrentVersion` returns `MAX(version)` and migrations are
forward-only, so any datastore that already ran the other branch's migration 10
will permanently skip the winner's and sit there looking healthy. Recreating
the reused CI database, the demo stack and any preview environment is the only
reliable fix, and that is a manual step at merge time rather than something
this PR can carry.

Post-rebase verification: alerter 71.1% against a ratchet raised 69 to 70,
collector 88.4% against 86 to 87, both `golangci-lint` clean, all tests passing.

Issue #453, opened 2026-09-14
Fix the 575 client typecheck errors and gate them in CI


dpage:
`npm run typecheck` (`tsc --noEmit`) in `client/` reports 575 errors on `main` (as of 2026-09-14, commit a80a2946), and nothing runs it: the script exists in `package.json` but neither the Makefile nor `.github/workflows/ci-client.yml` invokes it. The tsconfig has `strict`, `noUnusedLocals` and `noUnusedParameters` enabled, so the errors are what the JavaScript-to-TypeScript conversion left behind rather than a configuration problem.

About 247 of the 575 are in test files. The bulk falls into four groups:

- **Implicit `any` (about 206):** unannotated parameters, destructured props and variables, e.g. the `blackout` and `onStop` props in `BlackoutPanel.tsx` and callbacks in `EffectivePermissionsPanel.tsx`.
- **Property does not exist (77):** values typed as `{}` or too narrowly and then indexed, e.g. `options.xAxis` in `Chart/Chart.tsx`, which alone accounts for 25.
- **Unused declarations (59):** mostly `import React` left in test files that no longer need it under the automatic JSX runtime.
- **Test-only type mismatches (about 100):** `vi.fn()` mocks passed as typed props (worst in the `ServerDialog` and `ClusterNavigator` tests), `MockInstance` given too many type arguments after a Vitest upgrade, and `Expected 1 arguments, but got 0` where mocked handlers are called bare.

The remainder is a scatter of nullability, `Cannot find name` and index-signature errors, concentrated in `EventTimeline/`, `StatusPanel/` and the admin panel.

## Proposed approach

1. Work the count down to zero directory by directory, with the test files as the cheapest first tranche. Fix types properly rather than sprinkling `any` or `@ts-expect-error`.
2. Once at zero, add `npm run typecheck` to the client CI job and to the client `make lint` (or `make test-all`) target so it cannot regress.

Reproduce with:

```bash
cd client && npx tsc --noEmit 2>&1 | grep -c "error TS"
```

PR #460, opened 2026-09-14
Give the AdminTokenScopes test file a load-tolerant timeout

### LINKED ISSUE #422: AdminTokenScopes clipboard test times out under full-suite parallel load
Noticed independently by two pieces of work today, and confirmed to be
pre-existing rather than introduced by either.

## Symptom

The test at
`client/src/components/AdminPanel/__tests__/AdminTokenScopes.test.tsx:256`,
"copies the token and swaps to the Copied! state on success", times out at
Vitest's default five-second limit when the full client suite runs in parallel.
It passes reliably in isolation, and it also passes in the full suite with a
raised timeout, for example `--testTimeout=20000`.

CI's client jobs pass on both Node 22 and 24, so this does not currently block
anything. It bites developers running `make coverage` or the full suite locally,
where it produces a failure that has nothing to do with whatever they were
working on. That is corrosive in a small way: a suite that cries wolf trains
people to ignore it.

## Why it is timing out

The test waits for the component to swap into its "Copied!" state after a
clipboard write. Under parallel load the machine is contended enough that the
transition does not land inside five seconds. The helper driving it is
`clickCopyAndAwaitCopiedState` at line 172 of the same file.

## Suggested fix

Raising the timeout for this test, or for the file, is the cheap remedy and is
probably sufficient. It is worth a moment's thought first about whether the
assertion can instead await a deterministic signal rather than a state
transition that races the event loop, since a raised timeout treats the symptom
and a deterministic wait removes the race. Either is an improvement on the
current position.

dpage:
## Summary

The copy-to-clipboard test in `AdminTokenScopes.test.tsx` timed out at Vitest's five-second default whenever the full client suite ran under parallel load, whilst passing in isolation. The cause is not the clipboard wait itself: several tests in that file drive the whole create-token flow through `userEvent` first and take two to three seconds each on an idle machine, so a load-induced doubling breaches five seconds.

- `vi.setConfig({ testTimeout: 20_000 })` at the top of the file raises the ceiling for this file only; a genuine hang still fails.
- The copy helper asserts the clipboard call and the copied state in a single `waitFor`, so a second polling cycle cannot race the component's two-second revert timer.
- `ServerDialog.test.tsx`, `AdminEmailChannels.test.tsx` and `AdminWebhookChannels.test.tsx` drive multi-step dialogs the same way and timed out in the same loaded run, so they get the same file-level ceiling.

## Test plan

- [x] The four files in isolation: 166 tests pass; lint clean
- [x] `cd client && make test-all` under heavy load (load average 24 to 48): 4025 tests pass

Closes #422

## Summary by CodeRabbit

- **Tests**
  - Improved test reliability for administrative token, email channel, webhook channel, and server dialog workflows.
  - Extended test timeouts to better support multi-step interactions under parallel execution while still detecting genuine hangs.
  - Reduced timing-related failures in clipboard-copy verification by coordinating copied-state checks within a single polling cycle.


AntTheLimey (approved):
**Approving. The fix is correctly scoped, and the clipboard change is a
real race fix rather than a wider timeout papering over one.**

**→ Verified `vi.setConfig({ testTimeout: 20_000 })` is file-scoped**
and does not leak to siblings in the same worker, so a genuine hang
elsewhere still fails at 5000ms. All four files apply it identically and
no production code changed.

**→ The merged `waitFor` in `clickCopyAndAwaitCopiedState` closes a
real gap.** Checked against `AdminTokenScopes.tsx`, which reverts the
copy icon 2000ms after a successful write, so collapsing two
independent polling loops into one removes an actual race rather than
hiding it.

**→ Measured here, not taken on trust:** the four touched files pass
166/166, the full suite passes 4025/4025, lint is clean, and typecheck
reports 563 errors on both this branch and main, so this adds nothing to
the count tracked in #453. No application source changed, so the
coverage ratchet cannot move.

**→ One for a separate change, not this one: `AdminUsers.test.tsx` is
the most likely next file to reproduce #422.** Its `submits the create
call when the form is valid` test drives a multi-field dialog through
`userEvent.type` the same way, and a full-suite run here with no added
load clocked it at 3419ms, with two siblings at 2199ms and 2352ms, so it
is already at 68% of the 5000ms default. It passed under the load
average of 24 to 48 your test plan describes, so nothing is broken
today, but the same one-line ceiling would close it off.

dpage:
@coderabbitai review

PR #461, opened 2026-09-14
Detect a missing collector config file through the wrapped error (#421)

### LINKED ISSUE #421: Collector config-file error branches are unreachable because os.IsNotExist does not unwrap
Found whilst working on #389 and deliberately left alone there, since it was
outside that change's scope.

## Problem

`loadConfiguration` in `collector/src/main.go` tests the error returned by
`config.LoadFromFile` with `os.IsNotExist`, at lines 196 and 200. That error
has already been wrapped, at `collector/src/config.go:82`:

```go
return fmt.Errorf("failed to read config file: %w", err)
```

`os.IsNotExist` does not unwrap, so it returns false for the wrapped error
however the underlying failure arose. Two branches are therefore unreachable:

- Line 196, the explicit-path case, which is meant to report "specified config
  file not found: %s" and instead falls through to the generic "failed to load
  config file" message.
- Line 200, the auto-discovered-path case, which is meant to log that the file
  has vanished and fall back to compiled-in defaults, and instead falls through
  to the branch that treats the failure as unexpected.

The practical effect is mostly cosmetic for the first branch, since both paths
are fatal and the operator still gets an error naming the file. The second is
more meaningful: a genuine TOCTOU race, where an auto-discovered file is
removed between the helper's stat and the load, no longer falls back silently
as intended.

This is also why those blocks show as uncovered in the coverage profile; they
cannot be reached from a test either.

## Suggested fix

Use `errors.Is(err, fs.ErrNotExist)`, which does unwrap, in both places. Whilst
there, it is worth checking whether any other `os.IsNotExist` or `os.IsExist`
call in the tree is applied to a wrapped error, since the same trap applies
wherever `%w` has been used.

Tests for both branches become possible once the check is corrected, and should
be added in the same change to bring the touched unit up to the project's
coverage floor.

dpage:
## Summary

`loadConfiguration` tested the error from `Config.LoadFromFile` with
`os.IsNotExist`, but `LoadFromFile` wraps the `os.ReadFile` error with `%w`
and `os.IsNotExist` does not unwrap, so it never matched and two branches
were unreachable: the one reporting an explicitly specified config file as
not found, and the one falling back to compiled-in defaults when an
auto-discovered file vanishes between discovery and load. Both now use
`errors.Is` with `fs.ErrNotExist`.

Discovery goes through a package-level variable so a test can interpose
between the helper's stat and the load, which is the only deterministic way
to reach the vanished-file branch. The existing explicit-path test asserted
only that some error came back, which is why the defect survived it, so it
now asserts the message.

As the issue suggested, I checked the other `os.IsNotExist` call sites in
the tree: the server's `LoadConfigDataDir` and `pkg/fileutil` both test an
unwrapped `os.Stat` error directly and are correct as they stand.

## Test plan

- [x] `make test-all` in `collector` against local Postgres: green, lint
      clean, total coverage 88.3% against a floor of 86
- [x] `loadConfiguration` coverage 90.9% to 100%

`COVERAGE_MIN` stays at 86 deliberately. The CI matrix spans 87.1% to 88.0%
across Postgres 14 to 18 on `main`, so the ratchet sits below the matrix
minimum rather than below the local figure, and a 0.1 point lift does not
move that minimum.

Closes #421

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved collector configuration loading when configuration files are missing or disappear during startup.
  - Explicitly requested missing files now report a clear, path-specific error.
  - Automatically discovered files that vanish before loading now correctly fall back to compiled-in defaults.
  - Wrapped “file not found” errors are now handled consistently.

- **Documentation**
  - Added changelog details covering the configuration-loading fixes.


AntTheLimey (approved):
**Approving. This fixes a genuine piece of dead code rather than a
theoretical one.**

**→ Confirmed the root cause.** `LoadFromFile` wraps `os.ReadFile`'s
error with `%w`, so `os.IsNotExist` could never match and both
missing-file branches were unreachable. `errors.Is(err, fs.ErrNotExist)`
unwraps and is the right replacement.

**→ The error cases stay distinguished where it counts.** A permission
error or malformed YAML on an auto-discovered path is still fatal
rather than silently falling back to compiled-in defaults. That is the
branch that would have bitten an operator, and it is handled.

**→ The tests prove the defect.** Both drive the real wrap chain, one
through a genuinely missing path and one through the new
`discoverDefaultConfigPath` seam, and both fail against the pre-fix
code rather than asserting a constructed error.

**→ Measured, not assumed:** collector coverage 88.0% against the 86%
floor, with `loadConfiguration` and `LoadFromFile` both at 100%. gofmt
clean, lint 0 issues, `make test-all` green, changelog entry accurate.

**→ The 20 lines added to `testing-strategy.md` stay in scope.** They
record this repo's wrap trap and test seam, not general Go testing
advice, so they do not re-inflate what #444 removed.

PR #462, opened 2026-09-14
Stop treating missing metric data as a resolved alert condition (#407)

### LINKED ISSUE #407: Alerts flap, latch, or falsely resolve because missing data is treated as resolution

### LINKED ISSUE #454: Alerter engine integration tests refuse any database not named ai_workbench or postgres
The alerter's engine integration tests (`baselines_test.go`, `anomalies_test.go` and the other users of `assertLocalTestDSN` in `alerter/src/internal/engine/baselines_test.go`) fail with `refusing to run destructive integration tests against database ...; expected one of: ai_workbench, postgres` whenever `TEST_AI_WORKBENCH_SERVER` names any other database.

That makes it impossible to run several worktrees' suites concurrently on the dev host, because the only way to keep the Go suites from dropping each other's schema on the shared `ai_workbench` database is to give each run its own database (`ai_workbench_pr449` and so on). The collector and server suites are happy with that; the alerter's are not, so on 14 September the alerter engine tests were the one part of #394's verification that could not be run locally.

The loopback host check is the safety net that actually matters and should stay. The database allowlist adds little on top of it, so I'd suggest either dropping it, or relaxing it to a prefix match on `ai_workbench` so that per-worktree databases pass whilst the check still catches an env var pointing at an unrelated local database. The helper's doc comment and `CLAUDE.local.md` should be updated to match whichever is chosen.

dpage:
## Summary

The alert cleaner read a latest-metric query that returned no row for an
alert's connection and database, or no rows at all, as proof the condition
had resolved. Any gap in collection therefore cleared the alert and the next
sample raised it again. The registry now records per metric whether a missing
row means recovery, and clearing is additionally gated on the metric's
collector probe still reporting for that connection, so missing data and a
resolved condition are distinguishable in both directions.

`cache_hit_ratio` returned every delta row in its window with no ordering, so
the evaluator, which fires when any row violates, and the cleaner, which
stops at the first row, disagreed on identical data. It now reduces to the
newest qualifying delta per connection and database.

`slow_query_count` counted queries whose lifetime `mean_exec_time` exceeded
1000 ms, an average since the last statistics reset that never decays. It now
derives an interval mean from the change in total execution time over the
change in call count, differenced per statement identity rather than per
`queryid` alone, since each identity is a counter a reset can zero
independently.

The two slot metrics that had no freshness cutoff gained a fifteen minute
one, and the `pg_replication_slots.inactive` window widened from five
minutes, which equalled the probe interval and emptied whenever a collection
ran late, clearing a critical alert. A test now requires every registry entry
to bound `collected_at` or be allowlisted with a reason;
`pg_settings.max_connections` is the one exception, since its change-tracked
probe stores nothing whilst the configuration is unchanged.

`checkpoints_req_delta`, named in the issue, needed no change: #406 already
moved it to a one hour window that reports 0 rather than vanishing on a
single sample.

## The probe freshness gate

A security review of the first commit found that the new cutoffs had
introduced a fresh instance of the very bug this PR fixes. Every
clear-on-absence metric bounds the age of the samples it reads, so fifteen
minutes after a collector stops the query empties exactly as it does when the
condition ends, and the cleaner then reported a genuinely inactive
replication slot or runaway WAL retention as resolved, `AlertClear`
notification included. `replication_slot_inactive` is critical severity.

Each registry entry now names the collector probe that fills the table its
latest query reads, and an absent metric clears only whilst that probe is
reporting for the alert's connection, judged from the same
`GetProbeStalenessByConnection` data the staleness evaluator uses and bounded
at three collection intervals. The decision is a pure function, so a stalled
probe, a probe missing from the staleness view, an entry naming no probe and
a failed staleness read are all tested without a database; all four leave the
alert active and log where an operator will see it.

The trade-off is deliberate: because that view filters on `is_available`,
`is_enabled`, `is_monitored` and a non-null `last_collected`, an operator who
disables a probe or stops monitoring a connection keeps the alert until they
clear or acknowledge it. That is the safe direction. #465 covers the related
case of a probe going unavailable clearing its own staleness alert, which is
pre-existing and needs a decision about which reasons for absence are
legitimate.

## Test guard hardening

The same review found `assertLocalTestDSN` was called from two of eleven
alerter integration entry points; the other nine read
`TEST_AI_WORKBENCH_SERVER`, connected, and ran schema constants opening with
unconditional `DROP TABLE ... CASCADE` with no host or database check at all.
All fourteen reads now go through one helper per package, so a new
integration test cannot skip it. The database allowlist also admitted
plausible real names, so a tag must now contain a digit:
`ai_workbench_pr407` and `ai_workbench_issue407` pass whilst
`ai_workbench_prod`, `ai_workbench_live` and `ai_workbench_staging` are
refused. The loopback-only host check is unchanged.

## Test plan

- [x] `make test-all` in `alerter` against local Postgres: green, lint clean,
      total coverage 71.2%, with the ratchet raised from 69 to 71 (CI
      measures the same 71.2% on all five Postgres majors)
- [x] `TestAuditC8*` and `TestAuditC9*` inverted to assert the fixed
      behaviour, including interval-mean sub-cases for an idle statement, a
      statistics reset, a single sample and per-identity deltas
- [x] Engine tests for both missing-data paths and all four verdicts of the
      probe freshness gate
- [x] Slot freshness tests, including one proving the ten minute sample the
      old five minute window dropped is now kept
- [x] `TestMetricRegistryProbeName` checks every entry names a probe its own
      SQL reads; I separately confirmed all seventeen names are probes the
      collector actually configures
- [x] The one-sort rewrite of `slow_query_count` measured with `EXPLAIN
      (ANALYZE, BUFFERS)` on a seeded 24,000-row window: 68 ms against 93 ms,
      total cost 4289 against 5679, buffers identical
- [ ] CI green, Codacy clean (CodeRabbit has hit the organisation's review
      limit and refused an on-demand review, so it has not reviewed this PR)

## Notes for review

`cleanup_data_integration_test.go` is new because the 90% floor applies to
the touched `cleanup.go`, which was at 71.4% and is now at 97.4%.

The documentation pass found pre-existing inaccuracies it left alone; they are
now #466, and they overlap the section PR #455 is editing.

Closes #407
Closes #454


dpage:
@coderabbitai review

dpage:
@coderabbitai review

AntTheLimey (changes requested):
Two blockers, and both are the same mismatch: a freshness gate measured
in probe intervals sitting over windows written as fixed literals. One
item from my earlier pass is cleared by the rebase.

**→ `pg_node_role.subscription_worker_down` still has the exact defect
this PR fixes for `pg_replication_slots.inactive`.** Its window is
`NOW() - INTERVAL '5 minutes'` and the shipped `pg_node_role` probe
interval is **300 s**, so the window holds at most one sample and empties
whenever a collection lands later than the probe's own runtime.

**→ Your own changelog entry states the rule this breaks.** The slot
window widened "because five minutes equalled the probe interval and a
single late collection emptied the window and cleared the critical
`replication_slot_inactive` alert."

**→ The new gate does not catch it.** `probeFreshnessRatioLimit` is
**3**, so at a 300 s interval the probe reads as current until it is
**900 s** late, three times longer than the window it feeds. The cleaner
ticks every 30 s, lands in the gap, takes `ErrNoMetricData`, sees a
healthy probe and clears `subscription_worker_down`, which migration 1
seeds as critical and enabled. The resolve notification goes out and the
evaluator re-raises on the next sample.

**→ It is the only one of 16 `clearWhenAbsent` entries under three
intervals**, checked against the seeded intervals. Both documents this PR
adds already state the rule: `adding-rules.md` says "A metric whose probe
runs every 300 seconds therefore uses a 15 minute cutoff",
`metrics-queries.md` says "The window must be at least three probe
intervals", and `TestMetricRegistryLatestSQLFreshnessCutoff` only checks
that a cutoff exists. **Widen it to 15 minutes**, and consider having the
audit test compare the literal against the seeded interval.

**→ Second blocker: the gate is measured in configured probe intervals,
but every window is a hardcoded literal.** `classifyAbsentMetric` clears
at `StalenessRatio <= 3.0`, and that ratio comes from
`probe_configs.collection_interval_seconds`. The windows do not follow it.

**→ That is reachable without touching code.**
`docs/admin-guide/probes.md` tells operators to change that column. Set
`pg_replication_slots` to **600 s** and the 15 minute windows become
**1.5 intervals**, so ordinary collection empties them while the gate
stays quiet until 30 minutes, and `replication_slot_inactive` and
`replication_slot_retention_high` both clear and re-fire.

**→ A per-connection interval never reaches the ratio at all.** The same
page documents per-connection overrides, but
`GetProbeStalenessByConnection` joins
`probe_configs ... AND pc.connection_id IS NULL`, so the gate silently
judges against the global value.

**→ My suggested fix for both: gate on the metric's own window rather
than a multiple of the interval.** Clear only when the probe's
`last_collected` falls inside the window the latest query reads. That is
correct at any interval and removes the false-clear half of the first
blocker too.

**→ Optional: the widened allowlist regex admits numbered
production-shaped names.** The comment argues that "the production-shaped
names this guard exists to refuse are words", but `ai_workbench_prod2`,
`ai_workbench_live1`, `ai_workbench_v2` and `ai_workbench_demo9` all
carry a digit and pass. Loopback is still the real safety net, so this is
the comment overstating rather than a hole: anchor the tag to
`(pr|issue|sess)[0-9]+`, or say plainly that the digit rule is a
heuristic.

**→ Optional: `resolveAbsentMetric` reads
`GetProbeStalenessByConnection` once per absent alert.** The call sits in
`cleanResolvedAlerts`'s loop and returns the same slice every time.
`resolveExtensionGates`, directly above it since #409 landed, resolves
its lookups once per pass for exactly this reason.

**→ Optional: un-monitoring a connection leaves permanent critical alerts
that nothing explains.** `GetProbeStalenessByConnection` filters
`c.is_monitored = TRUE`, so decommissioning a server clears its
`metric_staleness` alert and makes every other alert on it
`absentMetricProbeNotReporting`. `DeleteOldAlerts` reaps only `cleared`
and `acknowledged`, so they sit active for ever with nothing beside them
saying why. An operator deliberately un-monitoring is a different signal
from a collector dying, and `connections.is_monitored` already tells them
apart.

**→ Optional: the `COVERAGE_MIN` comment still says 71.2% while the value
below it is 72.** `alerter/Makefile` describes the default as "pinned
just below the total measured on 14 September 2026 (71.2%...)" directly
above `COVERAGE_MIN ?= 72`, so as committed it reads as a floor pinned
above the measured total. Your commit message says 72.1% and I measure
72.1%. Worth correcting both figures.

**→ Optional: the two `test_dsn_guard_test.go` files are byte-identical
apart from the `package` line**, 112 lines each, and `6ec07eef` had to
edit both in lockstep. A small non-test package under
`alerter/src/internal/` would leave one copy and one
`TestAllowedTestDatabase`.

**→ The DSN item from my earlier pass is cleared, and #454 is genuinely
closed.** `6ec07eef` widens the regex and rewrites the knowledge-base
paragraph; I confirmed `issue407` and `sess2` pass while `prod`, `live`,
`staging`, `demo` and a bare trailing underscore are refused.
`ai_workbench_pr449` passes, loopback is untouched, and
`requireLocalTestDSN` is now the only place either alerter package reads
`TEST_AI_WORKBENCH_SERVER`.

**→ Everything else holds up, measured rather than assumed.** Neither new
query differences an aggregate: `slow_query_count` takes `LEAD` within
`(connection_id, database_name, queryid, userid, dbid, toplevel)` and
sums only afterwards, with the `CASE WHEN calls > prev_calls` guard
turning a `pg_stat_statements_reset()` into a NULL mean and a counted 0
rather than a negative delta, and `cache_hit_ratio` keeps `LAG` inside
`db_blocks` with `DISTINCT ON` outside it, collapsing evaluator and
cleaner onto the same row. Both registry accessors index a value-type
map, so an unknown metric gets `false` and `""` and cannot panic. All 31
entries name a probe, every name resolves to a seeded `probe_configs`
row, and both registry audits are registry-wide rather than spot checks.
On `70a1d261`: `make test-all` exit 0, **1228 tests and subtests
passing, 0 failures, 3 skips**; total coverage **72.1% against the 72
floor**, with `origin/main` at 71.1% in the same session; every touched
function at or above 90%, `classifyAbsentMetric` and `checkAlertResolved`
at 100%, `resolveAbsentMetric` at 93.3%; gofmt clean, golangci-lint 0
issues; changelog accurate against this head.

**→ Merge order: no constraint from my side.** #455 merged as `f6a016dc`
while this branch rebased, and the reconciliation is correct:
`deadlocks_delta` and `temp_files_delta` both carry
`probeName: "pg_stat_database"` and `TestMetricRegistryProbeName` passes
registry-wide. #464 and #468 conflict with you only in
`docs/changelog.md`, and #470 merges clean. #464 changes
`pg_stat_statements` column detection, but both query shapes already
select `calls` and `total_exec_time`, which is all the rewritten
`slow_query_count` reads, so they do not disagree semantically.

dpage:
Both blockers and four of the optionals are addressed in the seven commits from `9b5804da` to `451c5ba5`.

- `pg_node_role.subscription_worker_down` now reads a 15 minute window, and a registry audit fails any `clearWhenAbsent` entry whose window is under three of its probe's seeded intervals. The interval map that audit uses is itself checked against the collector's `probe_configs` seed, so the two cannot drift.
- The gate is now the metric's own window rather than a multiple of the interval. Each `clearWhenAbsent` entry declares `absenceWindow`, a test asserts it matches the shortest `collected_at` cutoff in its `latestSQL`, and `classifyAbsentMetric` clears only when the probe's `last_collected` falls inside that window. `probeFreshnessRatioLimit` is gone. Integration tests cover a raised global interval and a per-connection override, in both directions.
- Probe staleness is read once per cleanup pass, the `COVERAGE_MIN` comment and value are corrected from a fresh measurement (72.5% measured, floor 72.3), the DSN guard lives once in `internal/testdsn` with the tag anchored to `(pr|issue|sess)[0-9]+`, and `GetProbeStalenessByConnection` is at 100%.

The un-monitored connection point I have left for discussion: I agree `is_monitored` distinguishes the two cases, but deciding what an un-monitored connection's alerts should become (clear, acknowledge, or a distinct state) seems worth its own issue rather than a change here.

AntTheLimey (changes requested):
**→ Both blockers from the last round are properly fixed, and one new
one arrived with the optimisation.** `subscription_worker_down` is at
`INTERVAL '15 minutes'` against its 300 s probe with a matching typed
`absenceWindow`, and `probeFreshnessRatioLimit` is gone from `alerter/`
entirely — `classifyAbsentMetric` now gates on `entry.SinceCollected <=
window` and reads no interval column at all. The three audit tests are
the right shape, especially pinning the typed duration to the shortest
literal in the entry's own SQL.

**→ Blocking: `probeStalenessSnapshot` freezes `NOW()` for the whole
pass, so the effective window is `absenceWindow + passDuration`.**
`SinceCollected` is computed server-side once on first use and every
later alert in the pass is judged against that frozen value. Concrete
case: the collector has died, the slots probe last collected 4m58s
before the snapshot loads, the pass takes 10 s, and an alert judged near
the end is tested at a true 5m08s against a 5 minute window. The metric
query reads nothing, the snapshot says 4m58s, the alert clears and an
`AlertClear` goes out — #407's symptom, smaller. Ten of the sixteen
`clearWhenAbsent` entries use that 5 minute window. Fix is about three
lines: record `readAt := time.Now()` in `probeStalenessSnapshot.get` and
compare `entry.SinceCollected + time.Since(s.readAt) <= window`. Same
applies to `checkStalenessAlertResolved`, which compares a frozen
`StalenessRatio`.

**→ In fairness, the counter-argument is reasonable and you may want to
push back.** The exposure is one band `passDuration` wide at the end of
each window, every window is now at least three probe intervals, and a
short pass makes the error a fraction of a percent. It is raised as
blocking because the delta introduces it — `probeStalenessSnapshot` is
not on `70a1d261` — and because false clears on absent data are this
PR's entire subject.

**→ Optional: `TestSeededProbeIntervalsMatchCollector` skips rather than
fails when it cannot read the collector source.** A move or rename of
`collector/src/database/schema.go` turns it green for ever and
`seededProbeIntervals` quietly stops being checked against anything. The
`len(seeded) == 0` guard catches a shape change but not a path change.
Both trees are always checked out together here, so `t.Fatalf` is safe.

**→ Optional: one 118-character line in
`.claude/golang-expert/metrics-queries.md`** against the 79-character
wrap, ending "a 300 second probe and flapped on every late collection".
Everything else the delta adds is within 79.

**→ What passed.** `make test-all` exit 0 with 1249 tests, 0 failures;
coverage 72.5% against the ratchet you raised to 72.3, every touched
function at or above the 90% floor and the new `testdsn` package at
100%; `gofmt` and `golangci-lint` clean; all 13 CI checks green across
PG 14 to 18. The registry audit re-derives correctly — 16
`clearWhenAbsent` entries, 16 windows, none under three probe intervals.
The per-connection override is tested in both directions, and
`UpsertProbeAvailability` advances `last_collected` independent of row
count, so a recovered condition still clears. Neither knowledge-base
edit is a re-inflation; both replace text this PR makes false.

dpage:
Merged with both review rounds addressed; recording what changed, since the reviews were still standing at merge.

Round one's blockers were fixed earlier: `subscription_worker_down` widened to 15 minutes against its 300 second probe, and `probeFreshnessRatioLimit` removed so that `classifyAbsentMetric` gates on the metric's own window rather than a multiple of the configured probe interval.

Round two's frozen-clock blocker is fixed in `288015d6`: `probeStalenessSnapshot` records `readAt` before the datastore read and exposes `age()`, and both `classifyAbsentMetric` and `checkStalenessAlertResolved` now add that age, so the error always falls towards leaving an alert active rather than clearing it. Four of the five new tests were confirmed to fail with the age terms removed, with the fifth as a control.

The optional items are done too: `TestSeededProbeIntervalsMatchCollector` now fails rather than skips when it cannot read the collector schema, the over-long knowledge-base line is wrapped, and the `alerter/Makefile` ratchet and its comment agree at 72.5 against a 72.7% measured total. The un-monitored-connection case is deferred to #500.

PR #463, opened 2026-09-14
Untrack the committed collector binaries (#447)

### LINKED ISSUE #447: Two compiled collector binaries are committed to the repository and not gitignored
`collector/collector-linux` and `collector/collector-linux-amd64` are
compiled Linux binaries tracked in git. They are 13,756,441 and
13,768,225 bytes, about 13.1 MB each, and neither is matched by
`.gitignore`.

Both were added in 5a921058 ("Start the big redesign, with lessons
learnt from pgedge-postgres-mcp", 2026-01-05) and have been carried on
`main` since.

Why it matters:

- Every clone pays for roughly 26 MB of build output that the build
  produces anyway.

- Because neither path is ignored, a developer who builds in the
  worktree sees them as modified and can re-commit them without
  noticing. `git add -A` on the collector directory would do it
  silently.

- The tracked binaries are already stale against the source, so anyone
  who runs one is running January's collector.

Suggested fix: `git rm --cached` both paths, add them (or a broader
`collector/collector-*` pattern) to `.gitignore`, and confirm the
Makefile's build output location matches whatever pattern is chosen.
Rewriting history to drop the blobs is a separate decision and probably
not worth it.

Found while reviewing #442, which does not touch either file.

dpage:
## Summary

- Removes `collector/collector-linux` and `collector/collector-linux-amd64` from the index. They are 13 MB each, have been on `main` since January, and are stale against the source.
- Adds `collector/collector-*` to `.gitignore`, alongside the existing `collector/ai-dba-collector*` entries, so a local build cannot be re-committed by accident.

Neither path is produced by the build, so nothing else needed changing: the collector Makefile defaults to `collector/ai-dba-collector` and the root Makefile overrides `BIN_DIR` to `bin/`, and both were already ignored. History is not rewritten, as the issue suggests; the blobs stay in the pack.

## Test plan

- [x] `make -C collector build` still produces a binary at the expected path.
- [x] `git check-ignore` confirms both removed paths are now ignored.

Closes #447


AntTheLimey (approved):
**Approving. The untrack is done correctly and the ignore rule is broad
enough that #447 cannot come back under a different filename.**

**→ Verified it is an untrack, not a destructive rewrite.** Both
binaries are gone from the tree at head while their blobs still resolve
from history at 13,756,441 and 13,768,225 bytes.

**→ `.gitignore` carries the pattern `collector/collector-*`, not the
two literal names**, so a differently named future build cannot
recommit 13 MB by accident. It is scoped to one directory level, does
not reach `collector/src/`, and catches nothing else currently tracked.

**→ Nothing in the build depended on those files being present.** The
Makefiles, `collector/Dockerfile` and `ci-collector.yml` all build the
binary from source, so CI is unaffected. No Go code changed, so the
coverage ratchet is untouched. `make test-all` in `collector/` is green
at 87.7% against the 86% floor.

**→ Worth knowing, not blocking: the clone does not shrink.** The blobs
remain in history. That is a separate call for another day.

PR #467, opened 2026-09-14
Accept per-worktree databases in the alerter engine test guard

### LINKED ISSUE #454: Alerter engine integration tests refuse any database not named ai_workbench or postgres

dpage:
## Summary

- `assertLocalTestDSN` in `alerter/src/internal/engine/baselines_test.go` rejected any database not named exactly `ai_workbench` or `postgres`, so the engine integration tests could not run against the per-worktree databases (`ai_workbench_pr449` and so on) that keep concurrent suites from dropping each other's schema.
- The database name check is now a prefix match on `ai_workbench`, alongside the `postgres` default that CI's Docker image provides; the loopback-only host check, which is the guard that actually matters, is unchanged.
- Added `TestIsAllowedTestDB` covering the CI default, the shared dev database, a per-worktree name and three names that must still be refused.

`CLAUDE.local.md` has been updated locally to describe the per-worktree database convention; it is git-ignored, so it does not appear in this diff.

## Test plan

- [x] `cd alerter && make test-all` with `TEST_AI_WORKBENCH_SERVER` pointing at `ai_workbench_pr454`: passes, including the engine integration tests that previously refused to run.
- [x] Pointing the same variable at `postgresql://postgres@127.0.0.1:5432/postgres_unrelated` still fails fast with the refusal message.

Closes #454


AntTheLimey (changes requested):
**Blocking: the loosened guard now accepts `ai_workbench_prod`, which
this repo's own docs use as the example production database name.**

**→ `isAllowedTestDB` uses an unanchored `strings.HasPrefix`**, so
`ai_workbench_prod`, `ai_workbench_live` and `ai_workbench_staging` all
pass alongside `ai_workbench_pr449`. That name appears five times in
this repository, in both configuration guides and all three example
YAML files. Point `TEST_AI_WORKBENCH_SERVER` at a local copy of one and
the engine suite drops its schema, which is exactly the "env var left
pointing at an unrelated local database" case the comment says the
guard defends against. Please anchor the match with a regex rather than
a prefix test; #462 already carries one that does the job, so matching
it would save a reconciliation later.

**→ `TestIsAllowedTestDB` has no near-miss case**, so it could not have
caught this. Its three refusal cases are `production`, `""` and
`template1`, none of which shares the `ai_workbench` prefix. Please add
`ai_workbench_prod` as a name that must be refused.

**→ Everything else checks out.** The loopback host check is confirmed
unchanged and still doing the real defensive work, failing closed on
IPv6 loopback, resolved hostnames and unset hosts rather than only
matching the two literal strings it lists. The `postgres` branch is
still an exact match. The change is genuinely test-only, gofmt and
golangci-lint are clean, `make test-all` passes 392/392, and coverage
is unaffected: the engine package measures 59.5% on both main and the
branch, and the subproject total is 70.1% against the 69% floor.

**→ Worth settling before either lands: #462 rewrites this same
guard.** It replaces `assertLocalTestDSN` with a shared
`requireLocalTestDSN` helper using the anchored regex this review is
asking for, including a test that explicitly refuses
`ai_workbench_prod`. The two conflict textually in `baselines_test.go`
(confirmed with `git merge-tree`) and solve the same problem, so it is
worth deciding which one actually lands rather than fixing both.

dpage:
Both findings are correct, and rather than fix them here I'm closing this in
favour of #462, which already carries the guard you asked for.

I confirmed the `ai_workbench_prod` problem: the name appears five times on
`main`, in `docs/getting-started/configuration/collector.md`,
`docs/getting-started/configuration/server.md` and all three example YAML
files, and the unanchored `strings.HasPrefix` accepts it. The test genuinely
could not have caught it, since none of its three refusal cases shares the
prefix.

#462 is a strict superset of this PR rather than an alternative to it. Its
`allowedTestDatabase` regex anchors the match and requires the tag to hold at
least one digit, so `ai_workbench_prod`, `_live`, `_staging` and `_demo` are
all refused and pinned as refusals in `TestAllowedTestDatabase`; and it
replaces `assertLocalTestDSN` with a shared `requireLocalTestDSN` used by
every integration entry point in both the `database` and `engine` packages,
not just `baselines_test.go`. There is nothing here that it lacks, so fixing
this branch would only produce the reconciliation you flagged.

Closing as superseded. #462 will close #454 as well.

Issue #402, opened 2026-08-11
Rate derivation defects in the metrics query layer will corrupt derived rates


dpage:
## Description

The server's `<column>_per_sec` facility
(`server/src/internal/metrics/query.go`) is the intended fix path for
the charts that currently plot raw counters. Before that migration
happens, several correctness defects in the derivation need
addressing, otherwise the resulting rates will be subtly wrong in ways
that are hard to spot on a chart.

## Sum-then-delta breaks on dimension churn

`BuildDerivedMetricsQuery` sums a column across rows and then applies
`LAG()` to the sum. This breaks whenever the row set changes between
samples: a table created between samples inflates the sum into a
spike, and a table dropped produces a negative delta that is silently
discarded, losing a real data point.

The impact is worst on `pg_stat_statements`, where the collector takes
the top 1000 by execution time
(`collector/src/probes/constants.go:62`) and top-N membership churns
constantly.

Fix: delta per dimension key first, then sum the deltas — partitioning
by `schemaname, relname`, by `queryid, userid, dbid`, or by interface
name as appropriate. This is strictly more correct, and it localises a
counter reset to the affected row rather than poisoning the whole
series.

## Counter resets need the `stats_reset` guard

The existing negative-delta guard is correct but incomplete. If a
counter climbs back past its previous value within one sample
interval, the delta is positive but wrong; this is plausible at a 600
second interval on a busy system.

The collector already stores `stats_reset` for `pg_stat_database`,
`pg_stat_wal` and the checkpointer. The robust rule is to NULL the
interval whenever `stats_reset` differs between two samples,
regardless of delta sign.

Note that a clean restart preserves counters on PostgreSQL 15 and
later, since statistics are written to disk on shutdown and reloaded;
only a crash, immediate shutdown, restore or promotion discards them.
That is a further reason to key off `stats_reset` rather than
inferring resets from restarts.

`pg_stat_statements` needs particular care: entries are evicted when
the table is full and restart at zero on reappearance, with no
per-entry reset marker. The collector does not currently capture
`pg_stat_statements_info.stats_reset`; it should.

## LOCF is applied to rates, where it is wrong

`scanSeriesRows` applies last-observation-carried-forward to NULL
buckets uniformly. For gauges this is correct, since the last
observation remains the best estimate. For rates it is wrong: a
missing bucket means unknown, and LOCF will paint a flat plateau of
the pre-outage rate across an entire collector outage.

[...]

PR #468, opened 2026-09-14
Attribute a client address to a query in the drill-down (#384)

### LINKED ISSUE #384: Collect pg_stat_activity.query_id so client IP can be attributed to a query
Split out of #350, where surfacing the client IP / hostname for a query turned out not to be achievable without a collector change.

## Problem

`client_addr` and `client_hostname` are only available in `pg_stat_activity`, but our probe does not collect `query_id`, and `metrics.pg_stat_activity` has no such column. There is therefore no join key between activity rows and `metrics.pg_stat_statements`, so a client address cannot be attributed to a query id. Matching on query text instead would be unreliable and could attribute the wrong client, so it was rejected.

## Proposed change

- Add `query_id` to the `SELECT` in `collector/src/probes/pg_stat_activity_probe.go`.
- Add a `query_id BIGINT` column to `metrics.pg_stat_activity` via a new schema migration (next free version is 8; see the migration pattern in `collector/src/database/schema.go`).
- Join `activity.query_id = statements.queryid` to surface the client address in the query drill-down.

## Caveats worth documenting

`pg_stat_activity.query_id` exists only on PostgreSQL 14 and later, and is populated only when `compute_query_id` is enabled, so the probe needs a version and setting guard with a graceful fallback on older or unconfigured servers.

Even once collected, the association is inherently best-effort: `pg_stat_activity` is a point-in-time view, so a client address is only ever captured for statements that happened to be in flight when a snapshot was taken. The UI should present this as "last observed client" rather than implying it is the only client that ran the query.

dpage:
## Summary

- The `pg_stat_activity` probe now collects `query_id`, which schema migration 10 adds to `metrics.pg_stat_activity`. That is the join key the datastore was missing: without it there was no way to connect an activity row, which carries the client address, to a `pg_stat_statements` row, which carries the query identifier.
- `GET /api/v1/metrics/top-queries` joins the activity samples of the hour before the latest statements snapshot and returns `client_addr`, `client_hostname` and `client_observed_at`.
- The query drill-down on the object dashboard shows this beside the database user, as Last Observed Client.

The attribution is best effort, and the documentation says so rather than implying otherwise: `pg_stat_activity` is a point-in-time view, so a client is only ever captured for a statement that happened to be in flight when a sample was taken, and the value names the client seen most recently rather than the only one that ran the query.

Three details worth a reviewer's attention:

- **Version and setting guard.** `pg_stat_activity.query_id` exists only on PostgreSQL 14 and later, so older servers select a `NULL::bigint` placeholder to keep the column list stable. On 14 and later the column is itself NULL whenever `compute_query_id` is off or the backend is idle, so no separate check of that setting is needed; such servers simply resolve nothing.
- **The join is keyed on three columns**, not just `query_id`. `pg_stat_statements` is keyed on `(userid, dbid, queryid, toplevel)`, so one identifier can belong to rows in several databases and roles; keying on the identifier alone would let a query in one database be credited to a client that only ever connected to another.
- **A Unix-domain-socket backend reads as `local`**, following the convention the connection-groups query already uses, which leaves `client_observed_at` as the only reliable signal that nothing was observed at all.

A security review raised those last two points along with a query-identifier push-down into the new CTE, so that the drill-down's single-query lookup no longer pays a full window sort; all three are applied here.

## Test plan

- [x] Collector probe tests cover both version branches and assert the stored `query_id` matches what `pg_stat_activity` reported for a live backend.
- [x] Migration 10 has a column, comment and idempotency test.
- [x] Server tests cover newest-sample-wins, the per-database and per-role keying, the socket case, the never-observed case and one row per query identifier.
- [x] Client tests cover all four rendered states and the tooltip.
- [x] Go, client and OpenAPI gates pass locally; coverage floors held.

Not verified in a browser: the dev server was not running, and the value needs collected activity data to show anything but "Not observed".

Closes #384


dpage:
@coderabbitai review

dpage:
@coderabbitai review

AntTheLimey (changes requested):
Two blockers, and the first one means the feature currently does nothing
on any collector that is not a fresh install.

**→ The migration is numbered 10, but main is already at 11.** #457 took
10 for `stats_reset` and #455 took 11, both after you cut this branch.

**→ The collision fails silently rather than loudly.** `Migrate` reads
`currentVersion` once and skips any migration at or below it, so on an
upgraded collector this one hits `10 <= 11` and is skipped, logging
nothing beyond "Schema is up to date".
`metrics.pg_stat_activity.query_id` is never created.

**→ Then both consumers fail every cycle.** The probe's `Store` inserts
into a column that does not exist, and `buildTopQueriesSQL`'s new
`last_client` CTE selects the same missing column, so the top-queries
endpoint falls back to its empty response.

**→ Renumbering to 12 reaches further than `schema.go`.** Rename
`migration_v10_test.go` to `migration_v12_test.go`, since it collides
add/add with main's file of that name; update `schema_test.go`'s
`expectedVersions` to `{1..12}`; and correct **six** prose references to
"migration 10" across `docs/changelog.md`,
`docs/admin-guide/api/metrics.md` (two), the collector probe reference,
the collector schema doc and `.claude/golang-expert/metrics-queries.md`.

**→ Second blocker: the new `formatTimestamp` test fails at any negative
UTC offset.** In `ObjectDashboard/__tests__/types.test.ts`, "formats a
valid timestamp with date and time parts" renders
`formatTimestamp('2026-01-01T00:05:00Z')` and asserts `/2026/`, but
`formatTimestamp` renders in local time, correctly and by design, so west
of UTC that instant is 31 December 2025.

**→ Scope it before you try to reproduce it: CI passes and so does a UK
machine.** Measured: `America/New_York` gives `12/31/2025, 7:05:00 PM`
and `America/Chicago` `12/31/2025, 6:05:00 PM`, both failing, while
`UTC`, `Europe/London` and `Asia/Tokyo` all pass. **It will look green to
you and be red for every US-based engineer**, and it blocks `make
coverage` for them entirely. Nothing pins `TZ` in `.github/workflows` or
`vitest.config.js`.

**→ The fix, measured rather than assumed.** Move the fixture off the
year boundary to a UTC midday time such as `2026-06-15T12:05:00Z`. Both
that assertion and the `\d{2}:\d{2}:\d{2}` one hold in every zone from
`Pacific/Midway` (UTC-11) to `Pacific/Kiritimati` (UTC+14).

**→ Optional: take main's `types.test.ts` wholesale rather than merging
the two copies.** Main's arrived with #402 and is a strict superset,
covering null-bucket handling yours does not test. Yours encodes the
pre-#402 `extractLatestValue` semantics, where
`extractLatestValue([1, 7, 0])` expected `7` and now returns `0`, so a
textual merge reintroduces a wrong assertion.

**→ Optional: set the `vitest.config.js` ratchet from a post-rebase
measurement, not from the higher side.** Main carries
`91.0 / 89.8 / 87.0 / 78.71`; this branch measures **branches 78.6%,
0.11 points below main's 78.71% floor**, so carrying main's numbers
across unmeasured fails CI on branches immediately. #470 bumps to
`91.1 / 89.9 / 87.1 / 78.8`, which this branch also cannot reach.

**→ Everything else passed, and all three points you flagged are
correct.** The version guard needs no separate `compute_query_id` check,
because the column is itself NULL in exactly those cases; the join really
is keyed on database and role as well as the identifier, pinned by
`TestTopQueries_LastClientKeyedOnDatabaseAndRole`; and the
socket-as-`local` convention matches `connectionGroupsQueryByClient`. The
`last_client` CTE is a `DISTINCT ON` latest-value lookup partitioned on
`query_id, datid, usesysid`, not a delta, so the cross-entity
differencing problem does not apply. Collector coverage measures **88.0%
against the 86% floor** and server **59.7% against 57.4%**, both with
zero failures; `gofmt` and `golangci-lint` are clean on both; `make
openapi` regenerates with no diff and the OpenAPI test passes; the client
adds **0** new TypeScript errors to the 563 already on main (#453);
`QueryDetail.tsx` measures **100% lines**; and all four rendered states
plus three tooltip variants are genuinely covered. The label, tooltip and
user-guide copy are all honest that the attribution is best effort and
sampled.

**→ Merge order: six conflicts against main**, one of them the migration
pair above, which renumbering resolves. #462 and #464 do not touch the
client, and #464 does not conflict with you at all. #470 overlaps only on
the `vitest.config.js` ratchet, so whichever of main, this PR and #470
lands last must set it from an actual post-merge measurement.

**→ No visual pass this round**, because the branch cannot merge cleanly
yet and a production-build pass would be of a tree nobody ships. Worth
doing once the conflicts are resolved, since you flagged the feature was
not checked in a running dev server.

dpage:
Rebased onto `main` and both blockers are fixed.

- The migration is now number 12, with main's 10 and 11 untouched. The tests moved to `migration_v12_test.go` (main's `migration_v10_test.go` is kept as is), `expectedVersions` runs 1 to 12, and the prose references in the changelog, the collector schema and probe reference docs, the knowledge base and one handler comment follow. The two "migration 10" mentions in `docs/admin-guide/api/metrics.md` turned out to belong to #402's `stats_reset`, so they stay.
- The `formatTimestamp` fixture moved to `2026-06-15T12:05:00Z`, and the hour part of the time regex accepts one or two digits so a 12-hour locale cannot reintroduce the same class of failure. Verified under UTC, `America/New_York`, `Pacific/Midway`, `Pacific/Kiritimati` and `Asia/Tokyo`.
- Main's `types.test.ts` is taken wholesale, plus the one date-and-time test main lacks, so the stale `extractLatestValue` assertion is gone with it.
- Client coverage measured post-rebase at lines 91.05, statements 89.86, functions 87.07, branches 78.78, so main's thresholds already sit within 0.1 below every figure and stay as they are; taking main's test file back is what recovers the branch coverage you measured at 78.6.

Gates on the rebased head: server 21 packages passing at 60.1% against the 59.9 floor, `make openapi` no diff; collector `gofmt`, `golangci-lint` and coverage (87.7% against 87) clean; client lint clean and `tsc` at the 563-error main baseline. One caveat: the collector `scheduler` package peaks at 94 Postgres connections when run alone, so under `go test ./...` on an 8-CPU host it can trip `max_connections = 100`. It passes when run by itself, the same happens on `main`, and I will open a separate issue for it. The visual pass in a running dev server remains open.

AntTheLimey (approved):
**→ Both blockers from the last round are fixed, and the feature works in
the production build.** The column is at migration 12 with
`expectedVersions` running 1 to 12, every "migration 10" left in the tree
belongs to #402, and the handler comment you flagged in `fd7cc644` now
says 12. The `formatTimestamp` fixture is off the year boundary, and the
other fixtures in `types.test.ts` and `QueryDetail.test.tsx` are safe in
every zone. All four states — IP, resolved hostname, `local` socket and
never observed — render correctly in both themes, with no console
errors.

**→ Optional: the PR body still says "migration 10" twice.** The Summary
and the Test plan both do, while `schema.go` registers the column at 12.
Worth fixing before merge because main has its own migration 10 for
`stats_reset`, so anyone reading the merge record later goes looking in
the wrong place.

**→ Optional: at 400px wide a resolved hostname runs off the screen.**
With `review468-bench-client.review468_default (172.19.0.7)` the value is
clipped mid-character with no ellipsis, because
`SERVER_INFO_VALUE_BASE_SX` sets `whiteSpace: 'nowrap'`. Not blocking —
the token is main's, `Database User` already uses it, and the Query Plan
graph clips at the same width — but this is the longest value it has
met. `minWidth: 0`, `overflow: 'hidden'` and `textOverflow: 'ellipsis'`
on the new span would contain it, and the tooltip currently shows the
address and time but not the hostname, so it would want the full value
too.

**→ What passed.** Collector 88.2% against 87 and server 60.1% against
59.9, both ratchets unchanged; every touched function at or above 90%;
`gofmt` and `golangci-lint` clean; `make openapi` with no diff and 7/7
OpenAPI tests. The `DISTINCT ON (query_id, datid, usesysid)` join matches
`pg_stat_statements`'s own key, the PG 14 guard keeps `Store`'s columns
stable, and the new label and value clear AA contrast in both themes on
the existing tokens.

Issue #480, opened 2026-09-16
Deleting a connection has a 10 second budget for a cascade across every metrics table


dpage:
Noticed whilst adding a `statement_timeout` to the datastore pool in #387, and
left alone there as out of scope.

`deleteConnection` gives the whole operation a 10 second context deadline
(`server/src/internal/api/connection_handlers.go:520`). Inside that budget,
`DeleteConnection` (`server/src/internal/database/connection_queries.go:448`)
issues `DELETE FROM connections`, which cascades across twenty-odd `metrics.*`
tables, several of which are range-partitioned and hold the full retention
history for that connection.

Ten seconds is a plausible budget for deleting a connection that was added
yesterday, and an implausible one for deleting a connection that has been
collecting for the default seven-day retention, let alone a longer one. The
failure mode is not obviously benign either: the context cancel aborts the
statement partway, so the user sees a failure and the rows stay, and retrying
gets the same result at whatever size the history has reached.

Worth deciding what the operation should actually be. Options include giving
the delete its own, much longer deadline with an explicit
`SET LOCAL statement_timeout` to match; detaching and dropping the relevant
partitions rather than relying on `ON DELETE CASCADE` row by row; or making
deletion asynchronous, marking the connection gone immediately and reclaiming
storage in the background.

Pre-existing, and unrelated to the windowed top-queries work that surfaced it.

Issue #479, opened 2026-09-16
Collector migration list in the documentation is stale and misnumbered


dpage:
Found whilst documenting #387, and left alone there rather than widening that
change.

`docs/developer-guide/collector/schema-management.md` has a "Current
Migrations" list that documents migrations 1 to 5 and then jumps to 22 and 23,
whilst the collector's migrations actually run 1 to 12. So the list is missing
most of what exists, and the two entries at the end refer to numbers that do
not.

#387 inserted its migration 12 in numeric order and changed nothing else, so
the list is now internally consistent up to 12 and still carries the two
spurious entries after it.

Worth reconciling the whole list against `collector/src/database/schema.go` in
one pass, and deciding whether the document should enumerate every migration
at all: a list like this goes stale silently every time someone adds one, and
the code is the thing that is actually authoritative. A short description of
the migration mechanism plus a pointer to the source may age better than an
enumeration that nobody remembers to update.

Issue #478, opened 2026-09-16
Datastore pool defaults to four connections, so a few slow requests starve the API


dpage:
Found during the security review of #387 (PR for the windowed top-queries
endpoint), and separate from the mitigations landing there.

`PoolMaxConns` defaults to 4 (`server/src/internal/config/config.go:217`,
defaulted at `:1049`, applied in
`server/src/internal/database/connection_queries.go:227`). Every API endpoint
that touches the datastore draws from that pool, so four concurrent slow
requests occupy all of it and everything else queues: dashboards, the
connections list, the timeline, and the alerter's reads where it shares the
instance.

That is a low ceiling for a service whose dashboards fan out into several
concurrent metrics requests per page load, and it turns any one slow endpoint
into an availability problem for the whole API rather than for itself.

Two things worth weighing, not necessarily both:

- Raise the default. Four looks like a value chosen for a single-user
  development box rather than for a server rendering dashboards.
- Reserve capacity, or put a small concurrency limiter (a buffered-channel
  semaphore of two or three) around the expensive metrics aggregations
  specifically, so that the cheap endpoints keep a slot even when the
  expensive ones are saturated.

#387 caps the top-queries window at 30 days and sets a `statement_timeout` on
the pool, which bounds how long any one slot is held; neither changes how few
slots there are.

Issue #477, opened 2026-09-16
RBAC checker fails open when constructed with a nil auth store


dpage:
Found during the security review of #387 (PR for the windowed top-queries
endpoint). Pre-existing and unrelated to that change, but worth recording
because it is the foundation every handler's permission gate stands on.

`auth.RBACChecker.CanAccessConnection` returns `(true, AccessLevelReadWrite)`
unconditionally when `rc.authStore` is nil, and
`NewRBACCheckerWithSharing(nil, ...)` reaches that state silently. See
`server/src/internal/auth/access.go:188`.

Production wiring always supplies a real store, so I am not claiming an
exploitable path today. The concern is the shape: a fail-open default sitting
one constructor call away from every handler is the wrong default for an
authorisation primitive, and it is what currently lets the top-queries tests
pass without authentication, so the behaviour is load-bearing in the test
suite rather than merely latent.

Worth considering inverting it to deny, and having the tests inject a
permissive fake explicitly instead of relying on nil. That would make the
permissive case visible at each call site that wants it.

PR #481, opened 2026-09-16
Aggregate top queries over the selected time window (#387)

dpage:
`/api/v1/metrics/top-queries` had no time dimension at all: it pinned every
row to the latest `pg_stat_statements` snapshot, so the Top Queries
leaderboard ignored the dashboard's range selector and the Query Detail
overlay was half filtered, its charts honouring the selected range whilst its
headline statistics reported the most recent sample.

This is the first of two pull requests for #387, covering item 1. Items 2 and
3, the summary tiles and the `index_name` OpenAPI gap, follow separately and
will close the issue.

## What changed

The endpoint now takes `time_range` (default `1h`), `time_start` and
`time_end`, resolved through `metrics.ResolveTimeWindow` so its rejections
read identically to every other windowed endpoint.

Because `pg_stat_statements` exposes cumulative counters, the windowed figures
are summed deltas between consecutive samples rather than a snapshot reading.
That aggregation is not new: `queryStatsSQLTemplate` already did it for a
single `queryid`, and this generalises the same reviewed pattern to all of
them, partitioning the `LAG` by `(queryid, database_name, userid, dbid,
toplevel)` and discarding pairs whose deltas go negative at a counter reset.

## Behaviour changes worth a reviewer's attention

- **A statement collected but not executed during the window no longer
  appears.** That is the intended reading of "top queries in this window", and
  it applies to the existing presets, not only to custom ranges.
- **`min_exec_time` and `max_exec_time` remain lifetime values**, because
  lifetime extremes cannot be differenced. Ordering by either therefore sorts
  a windowed list on a lifetime figure; the order keys were kept rather than
  removed, which would have broken the API.
- **The Query Detail overlay lost its `Avg Time` tile.** `Mean Time` is now
  derived from the same window by the same arithmetic, so the two were showing
  one number twice. `/api/v1/metrics/query-stats` and its hook stay for API
  consumers.
- **The window is capped at 30 days on this endpoint**, tighter than the
  shared 366-day limit. See below.

## Performance

The aggregation reads every sample in the window rather than one snapshot. On
a 30-day fixture (8,640 collections, 2.6M rows) that cost about 22 seconds per
request before indexing, entirely an external merge sort.

Migration 12 adds a covering index, and the plan becomes
`WindowAgg -> Merge Append -> Index Only Scan` with no sort:

| Window | Before | After |
|---|---|---|
| 1h | 10.8 / 11.7 ms | 5.4 / 8.3 ms |
| 24h | 406 / 296 ms | 201 / 178 ms |
| 7d | 2.61 / 2.02 s | 1.43 / 1.50 s |
| 30d | 10.9 / 11.5 s | 5.0 / 5.5 s |

The index alone was not enough. Carrying the query text through the windowed
CTE made each row roughly 270 bytes rather than 85, so an index wide enough to
cover it would have been 889MB against a 1756MB table, and the planner
declined to use it anyway. The text is therefore resolved by a
`LEFT JOIN LATERAL` outside the `LIMIT`, once per returned row, through the
existing `idx_pg_stat_statements_object`, which is why that index stays.

Write cost of the new index: about +39% insert time (8.9ms to 12.4ms per
collector batch, against a 300-second interval), +56% WAL and +17% storage on
that table. The WAL figure is the one to watch on a busy datastore.

Migration 12 is a plain `CREATE INDEX` on a partitioned parent, so it takes
`ACCESS EXCLUSIVE` on every partition and blocks collector inserts for its
duration, about 15 seconds on the 2.6M-row fixture and proportionally longer
with a long retention.

## Security

Reviewed before opening. Two findings from that review are fixed here, both
created or materially worsened by this change:

- The window is capped at 30 days for this endpoint, applied after resolution
  rather than by tightening the shared `MaxCustomTimeSpan`, which other
  endpoints legitimately need at 366 days. Without it an authenticated user
  could request a 366-day window, twelve times the largest preset the client
  can produce, and with the CTE running twice per request and a pool that
  defaults to four connections, a handful of those starve the whole API.
- Query failures no longer report as an empty success. `42P01` keeps the empty
  200 for a workbench whose collector has never run; everything else now logs
  at `[ERROR]` and returns a 500, matching what `handleQueryStats` already
  does. Statement timeouts are a realistic outcome on this endpoint now, and
  were previously indistinguishable from "no data".

Also here, since it bounds the same risk: the datastore pool now sets
`statement_timeout` (new `database.statement_timeout`, default 30s), so a
runaway query is killed server-side rather than only cancelled client-side.

The review confirmed the injection-safety contract of `buildTopQueriesSQL`
still holds after the rewrite, and that the `LATERAL`'s use of the raw
`database_name` cannot cross a connection boundary.

Out of scope, filed separately: #477 (RBAC fail-open on a nil auth store),
#478 (pool sizing), #479 (stale migration list), #480 (connection delete
budget).

## Test plan

- [x] `make test-all` green from the repository root.
- [x] Server, collector and client coverage gates pass against their ratchets;
      none raised, since CI runs below local where the pgvector tests skip.
- [x] New integration coverage for a counter reset inside the window, a
      single-sample window, statements appearing and disappearing mid-window,
      the 30-day cap at its boundary, and every `ResolveTimeWindow` rejection.
- [x] `statement_timeout` proven server-side by running `pg_sleep` past the
      limit and asserting the pool slot is immediately reusable.
- [x] `mkdocs build --strict`.

One test was found passing vacuously and fixed:
`TestHandleTopQueries_SkipsUnscannableRows` widened `calls` to `text`, which
broke the summed delta so the statement failed and the assertion never reached
the scan loop it claimed to exercise.

## Conflicts

Expect a conflict with #468, which rewrites `perf_summary_handlers.go`,
`QueryDetail.tsx` and `openapi.go`. Whichever merges second carries it.


dpage:
@coderabbitai review

AntTheLimey (dismissed):
**→ Request changes: the window totals are wrong on any server with
`pg_stat_statements` in more than one database, and the database filter
does not isolate a database.** Three smaller fixes ride along, one of
them client-side.

**→ Blocking: each database stores its own copy of the same counter,
and `totals` sums the copies.** The probe is database-scoped but reads
the cluster-wide view, so one `(queryid, userid, dbid, toplevel)`
counter lands once per database with the extension, at the same
`collected_at`. `samples` partitions by `database_name` too, so the
copies survive to `totals`. On the migrated schema, a counter growing
10 calls per 5-minute sample and stored under `db_a` and `postgres`
returned **240 calls and 2400 ms for a 1h window whose true figures are
120 and 1200**. Main read one copy, so this is new, and the Query Detail
header shows the same inflated figures. **It shows in the production
UI**: with the extension in `ai_workbench` and `postgres`, the datastore
held two copies per sample of the pgbench `BEGIN`, both `dbid` 16384,
and Top Queries showed **3,041,105 calls against 1,548,953** from
`pg_stat_statements` seven seconds later. Keeping one copy per
`(queryid, userid, dbid, toplevel, collected_at)` and partitioning on
those four fixes it; a fixture with one counter under two
`database_name` values would hold it.

**→ Blocking: `database_name` filters after `totals` has summed every
`dbid`.** With `queryid` 2002 at 1,200 calls in one database and 12 in
another, **`database_name=db_a` returned 1,212**. Resolving `dbid`
through `db_names` inside `samples` and filtering there makes a filtered
request sum only that database, and also stops `db_b` returning nothing,
which main already did.

**→ Blocking: two statements this PR adds are false on its own head.**
The `maxTopQueriesTimeSpan` comment and the new `metrics-queries.md`
paragraph say the aggregation has **no statement timeout**, which this
PR adds as a 30s pool default. Both name
**`idx_pg_stat_statements_object`** as the migration 12 index
benchmarked at 30 days; migration 12 creates
`idx_pg_stat_statements_identity_time`. The knowledge-base text on the
identity `LAG` will need updating with the first fix too.

**→ Blocking: `QueryDetail.fetchQueryData` can show the wrong window.**
Its effect cleanup sets `isMountedRef.current` false and the next run
sets it straight back to true, so an in-flight request from the old
range still applies. Switching `30d` to `1h` lets the slow `30d`
response land last and overwrite. `TopQueriesSection.tsx` in this PR
already has the `requestIdRef` guard and its test; `QueryDetail` needs
both. It did not show on an 8-minute-old dataset, where 30d and 1h
return the same totals, so this one rests on the code.

**→ Optional: the custom-range picker allows 366 days and this endpoint
refuses anything over 30**, so Top Queries and Query Detail show
`span must not exceed 30 days` while every other panel works; in the
overlay the error replaces the Query Text, Plan and Statistics sections.
The help text this PR edits now says Top Queries follows a custom window
and still quotes 366 days, so it is the natural place to say 30.

**→ Optional: label `Total Calls` and `Total Time` as windowed.** They
sit beside `Min Time (All Time)` and `Max Time (All Time)` with no
qualifier, so the row reads as lifetime figures. In a window with no
runs the All Time tiles go to `--` as well.

**→ What passed.** Server **60.2%** against 59.9 and collector **88.5%**
against 87, both on pgvector 0.8.6-pg17; client lines **91.06%**
against 91.0, with `QueryDetail.tsx` at 99.12% and
`TopQueriesSection.tsx` at 98.10% of statements. Every new server
function at 100% except `handleTopQueries` at 96.8%. `gofmt`,
`golangci-lint` and `npm run lint` clean; typecheck at 563 errors on
both main and this head. `make openapi` gives no diff. Reset handling,
mid-window appearance and the 30-day cap behave as described, and on
864,000 rows the 30d page runs index-only in 4,051 ms. In the browser,
pagination resets to page 1 on a range change, an empty window shows the
empty-state message, the help panel wraps cleanly at 400px, and the
console shows no new errors.

**→ Merge order: migration 12 is also claimed by #468, #476 and #492.**
If #468 lands first this becomes 13, and both
`DELETE FROM schema_version WHERE version >= 12` lines in
`migration_v12_test.go` must change too, or they wipe and replay #468's
migration. #468 also collides in logic: its `last_client` join hangs off
the `pss` row this PR removes from `deduped`, and the page now lists
columns explicitly, so the second to merge must re-join on
`latest_sample` and add the client columns or the scan returns a 500.
No textual conflict with #470, #474 or #494.

PR #488, opened 2026-09-16
Format and staticcheck-clean pkg/, and gate it in CI

### LINKED ISSUE #423: Formatting and staticcheck backlog under pkg/
Surfaced by CodeRabbit on #419, where it was correctly identified as
pre-existing and outside that PR's scope. Filing it so the backlog is recorded
rather than rediscovered on every subsequent PR that happens to touch `pkg/`.

## Unformatted files

`gofmt -l pkg/` reports six test files:

```
pkg/connstring/connstring_test.go
pkg/crypto/password_test.go
pkg/datastoreconfig/datastoreconfig_test.go
pkg/hostvalidation/hostvalidation_test.go
pkg/logger/logger_test.go
pkg/worker/worker_pool_test.go
```

All six are on `main` today. Since the project requires `gofmt` on all Go
files, these should simply be formatted. Note that CI does not currently catch
them, which is the more interesting half of the problem: whatever gate is meant
to enforce formatting is not covering `pkg/`, and fixing the six files without
fixing the gate means they will drift again.

## Staticcheck findings

CodeRabbit additionally reported `QF1001` and `ST1005` against
`pkg/fileutil/fileutil.go`, around lines 147 to 148, which date from #267. The
`QF1001` finding is a De Morgan simplification of the negated conjunction in
the tilde-path guard. I have not independently confirmed the `ST1005` finding,
so whoever picks this up should run the linter and work from its actual output
rather than trusting this description.

Worth checking as part of the same change whether `staticcheck` runs over
`pkg/` at all in CI, for the same reason as the formatting gate.

## Scope

This is deliberately a tidying task with no behavioural component. It should
not change any logic; if a `staticcheck` fix appears to change behaviour, that
is a separate bug and wants its own issue.

dpage:
## Summary

- Ran the six drifted test files under `pkg/` (`connstring`, `crypto`, `datastoreconfig`, `hostvalidation`, `logger`, `worker`) through `gofmt`.
- Cleared the two `staticcheck` findings in `pkg/fileutil/fileutil.go`: `QF1001` (De Morgan on the tilde-path guard) and `ST1005` (the unsupported-tilde-path error ended in punctuation). Both confirmed with `golangci-lint` v2.5.0 on default config; neither changes behaviour, and nothing in the repo asserts the message text.
- Fixed the gate, which is the more interesting half. `pkg/` is its own Go module and was entirely ungated: no Makefile, no workflow, and every lint, vet and `gofmt` step in `ci-alerter.yml`, `ci-server.yml` and `ci-collector.yml` runs with `working-directory: <sub-project>/src`, so none of them ever looked at `pkg/`. This adds `pkg/Makefile` (modelled on `alerter/Makefile`) and `.github/workflows/ci-pkg.yml`, and points the root `Makefile`'s `test`, `lint` and `test-all` targets at the new Makefile rather than a bare `go test ./...`. Without that, the six files would simply drift again.

No coverage target or ratchet was added for `pkg/`; that would be a new gate the issue did not ask for, and belongs in its own change.

## Test plan

- [x] `cd pkg && gofmt -l .` prints nothing
- [x] `cd pkg && golangci-lint run ./...` reports 0 issues
- [x] `cd pkg && go vet ./...` clean
- [x] `cd pkg && make test-all` passes
- [x] `server/src`, `collector/src` and `alerter/src` all still build
- [x] `ci-pkg.yml` parses as YAML; no Postgres service needed, since no `pkg` test touches a database

Closes #423


AntTheLimey (approved):
**→ Approve: the diff is the cleanup it says it is, and the new gate
fails when it should.** The six test files are formatting-only, with
`git diff -w` empty on all six and every test-function count unchanged.
The one logic line in `ExpandTildePath` is a De Morgan rewrite with the
same truth table.

**→ Optional: `pkg/` has no `.golangci.yml`, so the new gate runs only
golangci-lint's defaults.** Collector, server and alerter each carry a
config that enables gosec; `pkg/` gets errcheck, govet, ineffassign,
staticcheck and unused. So the `// #nosec G304` in `fileutil.go` guards
nothing, and `pkg/crypto`, `pkg/connstring` and `pkg/hostvalidation`
get no security linting. A `pkg/.golangci.yml` matching the sibling
modules would close it.

**→ What passed.** An unformatted function made `make fmt-check` fail,
and the pre-rewrite De Morgan pattern made `make lint` fail with
`QF1001`. `pkg`'s `make test-all` is green with `-race`, the root
`test`, `lint` and `test-all` targets delegate into it, `ci-pkg.yml`
triggers on PRs touching `pkg/**`, and per-package coverage is identical
on main and this head. No conflict with #491, whose new
`pkg/embedding` files already pass this gate.

Issue #56, opened 2026-04-15
Alerter fails with foreign key violation when calculating baselines for deleted connections


imtiazqa:
When a monitored connection is deleted, the alerter continues attempting to calculate and upsert baselines for the deleted connection ID, resulting in foreign key constraint violations:

ERROR: insert or update on table "metric_baselines" violates foreign key constraint 
"metric_baselines_connection_id_fkey" (SQLSTATE 23503)

**Root Cause:**

The alerter's baseline calculation query retrieves connection IDs from the metrics data without first verifying that the connection still exists in the connections table. When a connection is deleted, orphaned metric data remains and the alerter tries to upsert baselines for a non-existent connection ID.

**Steps to Reproduce:**

- Add a monitored connection — it gets assigned ID 1.
- Allow the collector to gather metrics for that connection.
- Delete the connection from the workbench.
- Wait for the alerter's next baseline calculation cycle.
- The alerter logs foreign key violation errors for connection ID 1.

**Expected Behavior:**

The alerter should skip baseline calculations for connection IDs that no longer exist in the connections table, or the connection deletion should cascade and clean up all related metric data.

**Proposed Fix:**

**Either:**

- Add ON DELETE CASCADE to the metric_baselines_connection_id_fkey foreign key, or
- Filter out deleted connection IDs in the baseline calculation query before attempting upserts.

I am seeing following errors in logs 

```
[alerter] Calculating baselines for 2 connections, 28 rules (lookback: 7 days)
[alerter] ERROR: Failed to upsert 'all' baseline for pg_stat_activity.count on connection 5: ERROR: insert or update on table "metric_baselines" violates foreign key constraint "metric_baselines_connection_id_fkey" (SQLSTATE 23503)
[alerter] ERROR: Failed to upsert hourly baseline for pg_stat_activity.count hour 15 on connection 5: ERROR: insert or update on table "metric_baselines" violates foreign key constraint "metric_baselines_connection_id_fkey" (SQLSTATE 23503)
[alerter] ERROR: Failed to upsert hourly baseline for pg_stat_activity.count hour 14 on connection 5: ERROR: insert or update on table "metric_baselines" violates foreign key constraint "metric_baselines_connection_id_fkey" (SQLSTATE 23503)
[...]

PR #489, opened 2026-09-16
Prefer time-aware baselines and scope anomaly detection by database

### LINKED ISSUE #408: Anomaly detection ignores time-aware baselines, is disabled for 14 metrics, and mixes up databases

dpage:
## Summary

Anomaly detection always scored against the global `all` baseline, because `GetMetricBaselines` ordered by `period_type` alphabetically and the detector took the first row; the hourly and daily baselines were written every cycle and never read. Detection was also permanently disabled for the 13 registry metrics without a historical query, since the fallback baseline built from a single current sample could never pass the warmup gate, and per-database metrics were scored against an arbitrary database's baseline with a NULL `database_name` on the candidate.

- Baseline selection now prefers the hourly baseline for the current UTC hour, then the daily baseline for the current UTC weekday, then `all`; the first warm candidate wins and detection is suppressed when none is warm. Hour and weekday keys come from one shared helper on both the writing and selecting sides.
- Metrics without `historicalSQL` are skipped by baseline calculation and detection via `SupportsBaselines`; the fallback path is removed and leftover rows for unsupported metrics are swept each baseline cycle so the server's `get_metric_baselines` tool stops seeing permanently cold rows. Adding historical queries for those metrics is a follow-up.
- `GetMetricBaselines` takes the database name with NULL-aware matching; every latest value for a connection is scored against its own database's baseline and the candidate carries `database_name`, so deduplication is per database. `GetLatestMetricValues` runs once per rule rather than once per rule per connection.

The deadlocks and temp files latest-versus-historical mismatch was already fixed by #455. The replication slot pair (`inactive_count`, `max_retained_bytes`) compute the same per-sample statistic in both queries, so nothing changed there.

The audit tests from #410 for C6, C7 and C10 now assert the fixed behaviour and the `ALERTER_DEFECT_DEMO` variants in the engine package are gone. Alerter coverage rises from 71.1% to 71.9% locally and the ratchet moves to 71.

## Test plan

- [x] `cd alerter && make test-all` (fmt, tests, coverage, golangci-lint) against a local `ai_workbench_pr408` database

Closes #408


dpage:
@coderabbitai review

AntTheLimey (dismissed):
The daily tier you added to the preference order can never be selected
at the shipped defaults, so the documented three-step order is a
two-step order in practice. `calculateBaselines` passes
`Baselines.LookbackDays` (default 7) into `GetHistoricalMetricValues`,
every branch of which filters `collected_at > NOW() - INTERVAL '1 day' *
$1`, so `earliestTimestamp` is never older than 168 hours, and
`UpsertMetricBaseline`'s `DO UPDATE SET earliest_sample_at = $13`
overwrites it each cycle so no older value survives. `isBaselineWarm`
then demands `now.Sub(EarliestSampleAt) >= 336h` for a `daily` row,
which `Daily.MinSpanHours: 336` sets in `config.go` and in both
`examples/ai-dba-alerter.yaml` and `docker/config/ai-dba-alerter.yaml`.
169h < 336h, always.

The failure is silent, which is why I would rather it were fixed here
than found later. An operator on the shipped config gets `all` until day
5, hourly from day 5 on (`Hourly.MinSpanHours: 120` against the same
168-hour ceiling), and never a weekday baseline, with no log line saying
so, because `selectBaseline` falls through to a warm candidate without
reporting the cold one it skipped. The gate was invisible before this PR
only because nothing ever read a daily row.

Either raise `Baselines.LookbackDays` past 14 or lower
`Warmup.Daily.MinSpanHours` to 168 or less, and then bring the two doc
claims into line: `docs/developer-guide/alerter/anomaly-detection.md`
states the hourly/daily/`all` order with no caveat, and the changelog
entry promises the daily step. The selector logic itself is right and
needs no change.

`TestSelectBaseline` cannot catch this as written: `warmSince` is
`now.Add(-30 * 24 * time.Hour)`, an `earliest_sample_at` 30 days old
that `calculateDailyBaselines` cannot produce at a 7-day lookback, so
the `daily preferred over all` case passes against a row that will never
exist. Deriving `warmSince` from `LookbackDays` would have failed.

Two optional points. The rewritten Step 2 example in `adding-rules.md`
reads `FROM metrics.your_table WHERE collected_at > NOW() - INTERVAL '1
day' * $1` with no `JOIN connections`, and
`.claude/golang-expert/metrics-queries.md` (which this PR also edits)
makes that join mandatory for anything feeding `metric_baselines`;
`metric_baselines.connection_id` has `REFERENCES connections(id)`, so a
contributor copying the example reintroduces the #56 FK violation. And
anomaly alerts raised before this change carry `database_name` NULL,
which `GetActiveAnomalyAlert`'s NULL-aware predicate will no longer
match once the candidate carries a real name, so the first cycle after
upgrade adds a second alert for `cache_hit_ratio`, `deadlocks_delta` or
`temp_files_delta`; `cleanResolvedAlerts` only auto-clears
`alert_type = "threshold"`, so the old one sits there until someone
clears it. A sentence in the changelog would cover it.

Everything else measured clean. `make test-all` passes on `01a2ea0b`
with 423 top-level PASS lines and 2 intentional skips against a local
pgvector pg17; alerter total coverage is 71.9% against the new
`COVERAGE_MIN ?= 71`, up from a measured 71.1% on `origin/main`, and
every new or modified function clears the 90% floor, the lowest being
`GetHistoricalMetricValues`, `calculateHourlyBaselines` and
`calculateDailyBaselines` at 92.3%; `gofmt -l` is empty and
golangci-lint reports 0 issues. I reproduced the registry audit
independently and get your 31/18/13 split with the same 13 names. The
per-database scoping holds end to end: only `cache_hit_ratio`,
`deadlocks_delta` and `temp_files_delta` are `scanWithDB` among the
baselined metrics and all three have a matching per-database historical
scan, `metrics.pg_stat_database.database_name` really is per row
(`datname = current_database()` in the probe, and it is in the PK), and
every delta CTE partitions by `connection_id, database_name` before
summing, so there is no cross-database contamination or double-counting.
The unique index's `COALESCE(database_name, '')` keeps the new
per-database rows from colliding on upsert. The knowledge-base section
is all repo-specific fact, not general advice. There are no open review
threads: CodeRabbit was rate limited and Codacy reports 0 new issues.

On merge order: #494 conflicts with this branch textually, in
`.claude/golang-expert/metrics-queries.md` only, because both of you
append a Related Issues bullet immediately above `#409` and both add a
section elsewhere in the file. `docs/changelog.md` auto-merges. Keep
both bullets and both sections. There is no semantic disagreement and no
shared Go file, so either can land first.

AntTheLimey (approved):
The daily-tier blocker from round 1 is fixed, and I'd approve this as
it stands. `DefaultLookbackDays` is now 15, so
`EffectiveLookbackDays() * 24` gives 360 hours against
`Daily.MinSpanHours: 336`, a 24-hour margin rather than a boundary
reading, and `TestShippedWarmupTiersReachable` plus the new
`TestUnreachableWarmupPeriods` and `TestEffectiveLookbackDays` cases pin
that relationship directly rather than through an arbitrary 30-day test
fixture. `TestSelectBaseline` now derives its warm `earliest_sample_at`
from the shipped lookback, exactly what round 1 asked for, and its three
new subtests confirm the reported cold row on a fall-through. I also
like the follow-up commit: the changelog told operators to raise
`lookback_days` to "at least 14," but 14 days is exactly the daily
gate's 336-hour boundary, and the historical query's strict `>` filter
means a freshly refreshed row sits just under that boundary and would
flap between warm and cold — correcting the advice to 15 is right and I
would not have caught that myself without checking the inequality
direction.

Both round-1 optional points are resolved. The `adding-rules.md`
historical-query example now joins `connections`, closing the FK-violation
path. The changelog now states plainly that anomaly alerts raised before
this change carry no database name, may duplicate on the first
post-upgrade detection for `cache_hit_ratio`, `deadlocks_delta` and
`temp_files_delta`, and are not auto-cleared.

One optional item remains, not a blocker: the branch measures 72.1%
against `alerter/Makefile`'s unchanged `COVERAGE_MIN ?= 71`, up from
71.9% at round 1's head. The repo's ratchet rule says to raise it
whenever a change lifts the total; bumping it to 72 would close that
gap.

Everything else measured clean on the final head (`bfe97318`).
`make test-all` passes with 429 PASS lines, 0 failures and the same 2
intentional skips as round 1, against a local pgvector PG 17; every
function this round's commit touches clears 90% (the lowest at 96.2%);
`gofmt -l` is empty and golangci-lint reports 0 issues; and CI is green
across all four Alerter Postgres majors. Round 1's #494 conflict note is
moot now that #494 has merged, and against current `origin/main` this PR
merges with zero conflicts. Among the still-open PRs, #491 shares three
files with this one but is disjoint in function and merges clean; #462
and #476 collide trivially in the Makefile ratchet line and the
changelog's Fixed section; and #462 additionally collides for real in
`audit_defects_test.go` and `baselines_test.go`, since #462 still calls
the two-argument `GetMetricBaselines` this PR extends to three — whoever
rebases second onto the other needs to fold #462's new test bodies into
the new signature by hand.

**Note added at posting time:** this PR merged with one more commit on
top of what was reviewed, "Reconcile the #408 baselines work with the
#407 alerter changes" (`29bb3ba1`) — a 2-line rebase reconciliation
against #462 landing first, exactly the collision this review flagged
above. It removes a duplicate test guard `#407` had already moved into
the shared `testdsn` package. Mechanical, doesn't change the verdict.

PR #491, opened 2026-09-16
Remove the embedding model allow-list

### LINKED ISSUE #427: Impossible to use unrecognized embeddings with openai model provider.
This may be a more broad issue related to model provider embedding restrictions.

In `pkg/embedding/libprovider.go`, we list valid provider mappings for certain embeddings:

```go
var supportedEmbeddingModels = map[string][]string{
        "openai": {"text-embedding-3-large", "text-embedding-3-small", "text-embedding-ada-002"},
        "voyage": {"voyage-3", "voyage-3-lite", "voyage-2", "voyage-2-lite"},
        "gemini": {"gemini-embedding-001", "gemini-embedding-2", "gemini-embedding-2-preview"},
}
```

However, when using local llama.cpp or other local model servers that reproduce the OpenAI protocol, this means nominally functional embeddings are not allowed by this filter. The current workaround is to use any embedding model and simply label it in the model server as one of the supported options. For example, in a llama.cpp preset file:

```yaml
[text-embedding-3-small]
model = /models/nomic-ai/nomic-embed-text-v1.5-GGUF/nomic-embed-text-v1.5.Q8_0.gguf
embedding = true
```

Some options to consider:

1. Relax the embedding model restrictions (even OpenAI could add new embedding models in the future).
2. Allow user-specified validation override lists in the server configuration.
3. Poll the model server for supported embedding models as validation rather than an artificial gate.
4. Document the workaround as the official method for llama.cpp, vLLM, and other systems that operate through leveraging the openai provider.

dpage:
## Summary

- The embedding package rejected any model name outside a hard-coded
  per-provider list, so an OpenAI-protocol-compatible local model server
  such as llama.cpp or vLLM could not be used without relabelling the
  model in the server's preset, and a genuinely new provider model was
  blocked until the list was edited. That list, and the
  construction-time validation it drove, are now gone.
- The configured model name is passed to the provider unchanged. The
  per-provider defaults still apply when the model is left empty, and
  `PadTo` still rejects an embedding wider than 4000 dimensions when it
  is stored or used as a query vector, so the guard that actually
  protects knowledgebase vector compatibility is untouched.
- Documentation reframed accordingly: the per-provider model lists now
  read as the commonly used or provider-published models rather than an
  exhaustive permitted set, and the server configuration reference
  describes pointing `openai_base_url` at a local model server.

Rather than relaxing validation only for a custom base URL, the
allow-list goes entirely, because a static list ages badly whichever
endpoint it guards.

## Test plan

- [x] `cd pkg && go test ./...` passes; `pkg/embedding` at 100% statement
      coverage.
- [x] `cd alerter && make test`, `make lint` and `make coverage` pass;
      the coverage ratchet holds at 71.0% against a 70% minimum.
- [x] `TestNewEmbeddingProvider_GeminiInvalidModel` converted to
      `TestNewEmbeddingProvider_GeminiCustomModel`, asserting that a
      model name outside the old list is accepted and passed through.
- [x] New cases cover a local-server model name over a custom base URL
      and the default-model fallback for each provider.

Closes #427


dpage:
@coderabbitai review

AntTheLimey (dismissed):
**→ Request changes: with the allow-list gone, the alerter silently
truncates any model wider than 1536 dimensions, and the new changelog
entry says the opposite.**

**→ Blocking: the alerter cuts every embedding to 1536 before the
4000-dimension check can run.** `embeddingAdapter.GenerateEmbedding` in
`alerter/src/internal/llm/llm.go` calls `vec.Resize(emb32, 1536)`, which
copies the first 1536 values and drops the rest with no error or log.
`embedding.PadTo`'s rejection of anything over 4000 only ever sees 1536.

**→ Why it blocks here, when the resize is older than this PR.** The
allow-list was the only thing keeping the alerter on known models, and
the ones it allowed survive truncation reasonably. This PR's docs now
invite any model from an OpenAI-compatible local server such as
llama.cpp or vLLM. A **4096-dimension model that was not trained for
truncation** is silently cut to 1536 and its similarity search
degrades, while the new changelog line says a model "wider than 4000
dimensions is still rejected with a dimension error". That holds for
the server's `memory/store.go` path, **not for the alerter**.

**→ Either fix works.** Scope the changelog's 4000-dimension sentence to
the server and document the alerter's 1536 resize in `alerter.md`, whose
existing "rejects a model that exceeds 4000 dimensions ... rather than
truncating" sentence is already wrong on main. Or drop the resize so the
alerter goes through `PadTo` like the server, noting that existing
anomaly rows were stored at the truncated width.

**→ Optional: the anomaly and chat-memory vector searches do not filter
on `model_name`.** Both tables store it per row, but
`anomaly_queries.go` and `memory/store.go` rank across every model's
vectors. Switching models is now easier, and
`search_knowledgebase.go` already guards the same case.

**→ What passed.** No references to the removed allow-list remain
anywhere in the tree; `pkg/embedding` at **100%**; alerter total
**71.0%** against 70.0, measured with pgvector; `gofmt`, `go vet` and
`golangci-lint` clean on both packages; every added docs line within
79 characters. No conflict with #488 or #489, whose edits to the two
alerter developer-guide pages sit in different sections.

AntTheLimey (approved):
**Approve: the new commit removes the truncation this review blocked
on, and the docs now match the code.**

**→ The blocker is fixed.** `embeddingAdapter.GenerateEmbedding` in
`alerter/src/internal/llm/llm.go` no longer calls `vec.Resize(emb32,
1536)`; it returns the model's native width, and both call sites in
`anomaly_queries.go` (`StoreAnomalyEmbedding`, `FindSimilarAnomalies`)
run that through `embedding.PadTo(embedding, embedding.MaxDimensions)`
(4000) before it reaches SQL, propagating a `PadTo` error rather than
swallowing it. A 4096-dimension model now genuinely hits the
"rejected with a dimension error" path the changelog describes, and a
model between 1536 and 4000 dimensions is stored and searched at full
width instead of being silently cut.

**→ Docs now match.** The changelog's #427 entry scopes the
4000-dimension rejection correctly and adds that pre-existing anomaly
rows stored below 1536 keep their truncated width until retention ages
them out. `anomaly-detection.md`'s model table drops the false
"(resized to 1536)" annotations for Ollama and Voyage.

**→ New test covers the exact failure case.**
`TestEmbeddingAdapter_GenerateEmbedding_KeepsNativeWidth` checks widths
of 3, 1536, 3072 and 4096 dimensions, asserting the last component of
the vector survives (a truncating adapter would zero it) and the
result stays unit-normalized.

**→ Optional, unchanged from round 1: no vector search filters on
`model_name`.** `anomaly_queries.go` and `memory/store.go` still rank
across every model's zero-padded vectors with no constraint, unlike
`search_knowledgebase.go`'s documented guard for the same case. Not
touched by this commit; still worth a follow-up.

**→ What passed.** `pkg/embedding` at **100%** statement coverage
(unchanged); `alerter/internal/llm` at **95.2%** (up from 92.0%),
`NewEmbeddingProvider` specifically up to **88.9%** from 80.6% on a new
Voyage-provider test; alerter total coverage **71.1%** against the
70.0% floor (up from 71.0%), measured against a fresh
`pgvector/pgvector:0.8.6-pg17` container; `gofmt`, `go vet` and
`golangci-lint` clean on both touched packages; no test failures; every
added doc line within 79 characters. No new review threads since round
1, and no CodeRabbit thread exists to address.

**Note added at posting time:** this PR merged with one more commit on
top of what was reviewed, "Log Tier 2 embedding width failures at
default verbosity" (`a683d62e`). It moves the similarity-search and
store failure logs from debug to error level so an over-wide model
does not fail silently, adds a dedicated integration test, and corrects
an unrelated dimension figure in the developer guide. It is additive to
this fix, not a walk-back of it, and does not change the verdict above.

Issue #439, opened 2026-09-10
pg_stat_statements column detection assumes the view is in pg_catalog, so block timings are silently discarded


dpage:
The `pg_stat_statements` probe decides which query shape to use by looking
for its version-specific columns in `information_schema.columns` filtered
on `table_schema = 'pg_catalog'`. The extension's view is not in
`pg_catalog` on an ordinary install, so both checks return false and the
probe falls through to its "PostgreSQL 12 and earlier" branch on every
modern server.

The consequence is silent data loss rather than a failure: rows still
arrive, but all four block-timing columns are stored as NULL and
`toplevel` is hardcoded to TRUE.

## Cause

`checkHasSharedBlkTime` and `checkHasBlkReadTime` in
`collector/src/probes/pg_stat_statements_probe.go` both filter on
`table_schema = 'pg_catalog'`. `CREATE EXTENSION pg_stat_statements`
installs into the current schema, normally `public`, so neither check
matches and the `else` branch wins.

## Evidence

On the dev host's PostgreSQL 18:

```
extname            | nspname
pg_stat_statements | public

table_schema | count
public       | 52
```

and the resulting row from the real probe:

```
toplevel=true shared_blk_read_time=<nil> shared_blk_write_time=<nil> local_blk_read_time=<nil>
```

On RDS PostgreSQL 17.11 the same check returns `f`, all 49 columns are in
`public`, and the data being thrown away is real:

```
 shared_read_ms | shared_write_ms | local_read_ms
         260.66 |            0.81 |          0.00
```

PostgreSQL 17 is the case that matters most, since that is the branch the
`shared_blk_read_time` detection exists to serve.

## Fix

Drop the schema predicate and resolve the view through the search path, or
look it up via `pg_class`/`pg_extension` rather than
`information_schema.columns`. Either way the check should not assume where
the extension was installed. Worth covering with a test that creates the
extension in a non-default schema.

Found whilst investigating #435; independent of that bug and affects
self-managed installs equally.

PR #493, opened 2026-09-16
Key the feature-detection cache per database, and stop the silence

### LINKED ISSUE #435: Top Queries/pg_stat_statements is not populating for AWS RDS when Primary has a replica
I'm running into an issue where the "Top Queries"/ `pg_stat_statements` is not being load when adding an AWS RDS postgres instance which has a replica.  I've double-checked all permissions for my pgedge monitoring user to make sure the permissions are correct and the user can access the pg_stat_statements table within my database.  

This is occurring on any AWS instance consisting of a Primary with a Replica instance.  The collector logs are not showing any errors.

To reproduce:

1. Create an AWS RDS postgres instance 
2. Create a read replica
3. Create a monitoring user with pg_monitor role
4. Grant role connect to database and usage on schema
5. Create pg_stat_statements extension on database
6. Generate activity in the database
7. Confirm monitoring user can connect and query pg_stat_statements
8. Add the instances to pgedge to monitor and wait 
9. Confirm "Top Queries" is not populated.

### LINKED ISSUE #438: Feature-detection cache is keyed per connection, not per database, so Top Queries silently never populates
The collector's feature-detection cache is keyed on the monitored
connection's name and the check name only, so for a database-scoped probe
the first database probed decides the answer for every other database on
that connection.

`pg_stat_statements` is the only probe in the tree that is both
`databaseScoped: true` and a user of `cachedCheck`; all thirteen other
call sites are server-wide, which is why this shows up specifically as
Top Queries never populating.

## Cause

`featureCacheKey` in `collector/src/probes/base.go` carries
`connectionName` and `checkName` but no database, and the entry is only
cleared when the connection's pool is recycled. `executeProbeForAllDatabases`
probes `current_database()` first and then loops the remaining databases
from `pg_database`, passing the same `conn.Name` every time, so the
verdict from the connection's own database is reused for all of them.

Where the extension is per-database, as on RDS, a monitoring connection
pointed at a database without it (commonly `postgres`) therefore
suppresses collection in the database that does have it, permanently.

## Reproduction

Verified two ways. Locally against PostgreSQL 18, with the extension
created only in `t435_app`, driving the real probe with a single
connection name:

```
t435_default (no extension): rows=0
t435_app     (has extension): rows=0     <-- suppressed by the cached verdict
t435_app     (fresh cache key): rows=17  <-- same database, same moment
```

And on RDS PostgreSQL 17.11 with a primary and a read replica, following
the steps in #435 exactly:

| Monitored connection's database | `appdb` result | Top Queries |
|---|---|---|
| `postgres` (RDS default) | 0 rows, despite 141 present | permanently empty |
| `appdb` (has the extension) | 141 primary / 39 replica | populates |

`pg_is_in_recovery()` had no bearing: primary and replica behaved
identically, so replication is not involved despite how #435 presented.

## Fix

Add the database to `featureCacheKey` so each database gets its own
verdict. Whilst there, the silence needs addressing too: `probe.Execute`
errors are logged at `Debugf` in both branches of
`executeProbeForAllDatabases`, so a probe that collects nothing produces
no visible error at default verbosity, which is how this reached a user
rather than CI. Relatedly, `recordAvailability` reports
`extension '...' not installed` whenever `allMetrics` is nil, so an empty
`pg_stat_statements` is indistinguishable from a missing extension in
`metrics.probe_availability`.

Reported downstream in #435.

### LINKED ISSUE #440: Collector enumerates rdsadmin and logs a connection error for it on every cycle
On RDS, `rdsadmin` has `datallowconn = true`, so the collector's database
enumeration returns it, but `pg_hba.conf` refuses the monitoring user
outright. Every database-scoped probe therefore attempts it and logs an
error, on every collection cycle, on every RDS deployment.

## Cause

`getDatabaseList` in `collector/src/scheduler/scheduler.go` selects from
`pg_database` filtered only on `datallowconn = true AND NOT datistemplate`.
`executeProbeForAllDatabases` then calls `GetConnectionForDatabase` for
each name and logs at `Errorf` before continuing when that fails.

The `continue` is correct, so nothing is broken; the problem is purely
that a guaranteed, unfixable failure is reported as an error, which trains
readers to ignore collector errors. Given that #435 arrived as "the logs
show no errors", the signal-to-noise ratio here matters.

## Evidence

From RDS PostgreSQL 17.11, with a `pg_monitor` user:

```
FATAL: pg_hba.conf rejects connection for host "...", user "pgedge_mon", database "rdsadmin", SSL encryption (SQLSTATE 28000)
```

and the database list the collector sees on a stock instance is exactly:

```
postgres
rdsadmin
```

## Fix

Either skip databases the monitoring user cannot connect to by checking
`has_database_privilege(current_user, datname, 'CONNECT')` in the
enumeration query, which is the general fix and covers more than RDS, or
demote a connection refusal on an unreachable database to a debug or
one-shot warning rather than a per-cycle error.

The privilege check is the better of the two, since it also stops the
collector attempting databases it was never granted.

Found whilst investigating #435.

dpage:
## Summary

Top Queries never populated on RDS where `pg_stat_statements` lives in the
application database but the monitored connection points at `postgres`.
The collector's feature cache was keyed on the connection name and check
name only, so the verdict from the connection's own database was reused
for every other database on that connection, for the life of the process.
The database is now part of the key, taken from the pool's configuration
so there is no extra round trip.

The logs were clean throughout, which is the other half of the bug:

- `probe.Execute` failures in `executeProbeForAllDatabases` were logged at
  `Debugf`, so a probe collecting nothing was invisible at default
  verbosity. They are errors now, whilst a genuinely absent extension
  reports itself through the new `probes.ErrExtensionNotInstalled` and
  stays at debug.
- That sentinel also lets `recordAvailability` distinguish an absent
  extension from one that is installed but currently empty; both used to
  record `extension '...' not installed`.
- The database enumeration now filters on
  `has_database_privilege(current_user, datname, 'CONNECT')`, so `rdsadmin`
  is no longer attempted and refused on every cycle (#440), along with any
  other database the monitoring user was never granted CONNECT on.

#439, the other problem found whilst investigating #435, is left to PR #464.

## Test plan

- [x] New `TestCachedCheck_PerDatabaseVerdicts` and
      `TestPgStatStatementsProbe_PerDatabaseExtensionVerdict` both fail on
      the unmodified tree with the verdict being reused across databases,
      and pass with the change.
- [x] New scheduler tests cover the merge and classification helpers, the
      `probe_availability` rows for present-but-empty versus absent (both
      database-scoped and server-wide), and the CONNECT filter driven as an
      unprivileged role.
- [x] `cd collector && make test-all` passes: total coverage 88.5%, probes
      91.2%, scheduler 91.3%, linter clean.

Closes #435
Closes #438
Closes #440


dpage:
Codacy raised seven findings on this branch. Five are Opengrep `concat-sqli`
on the new tests, and I have marked them ignored with written reasons rather
than changing the code:

- One is a plain false positive: the flagged `QueryRow` in
  `probe_availability_extension_test.go` uses a static SQL literal with `$1`
  and `$2` placeholders, so nothing is constructed at all.
- Four are test-only DDL, marked `TestCode`. `CREATE DATABASE`, `DROP
  DATABASE` and `CREATE ROLE` cannot take a placeholder for an identifier,
  and every name involved is built in-test from a fixed prefix and a
  nanosecond timestamp. This is the same shape the existing
  `scheduler_integration_test.go` and `database/schema_test.go` fixtures
  already use.

The remaining two are Lizard complexity warnings on
`executeProbeForAllDatabases`, which this change pushed to complexity 24 and
115 lines. Those are fair, and a follow-up commit extracts the per-database
loop body to bring the function back under both limits, with no behavioural
change.

dpage:
@coderabbitai review

AntTheLimey (dismissed):
The database-per-cache-key fix and the `ErrExtensionNotInstalled`
sentinel are both correct and well tested, but two things need fixing
before this merges.

**→ `.claude/golang-expert/feature-detection-cache.md` misdescribes
`executeProbeForServerWide`.** It claims that function got the same
Debugf-to-Errorf change as `executeProbeOnDefaultDatabase` and
`executeProbeOnDatabase`. It didn't: its generic non-timeout error
branch is untouched, still `Debugf`. Every server-wide extension probe
(the nine `system_stats`-backed `pg_sys_*` probes) still swallows a
genuine execution failure at default verbosity — the exact silence
this PR's title is about, just left standing on one code path. Please
either extend the fix there or correct the doc.

**→ The same file also says an absent extension is recorded in
`metrics.probe_availability`; the table is unqualified
`probe_availability`** per `schema.go` and per this PR's own
`probe_availability_extension_test.go`, which queries it without a
schema prefix.

**→ `TestGetDatabaseList_SkipsDatabasesWithoutConnectPrivilege`
self-skips, so #440 has no working regression test.** The role it
creates has no password, and the connection to it over TCP hits
`scram-sha-256` and fails with SQLSTATE 28P01 — reproduced locally
against the same `pgvector/pgvector:0.8.6-pg16` image family
`ci-collector.yml` uses, which sets no `POSTGRES_HOST_AUTH_METHOD`.
Give the role a password and thread it through `rolePool`'s connection
string the way `f.rawPassword` already is for the admin pool.

**→ Optional: `executeProbeForConnection`'s availability switch still
keys "not installed" off `allMetrics == nil` rather than the new
`extStatus == extensionAbsent`,** so a genuinely erroring probe still
gets mislabeled as a missing extension. Pre-existing, but the type you
just added would fix it cleanly.

**→ What passed.** `gofmt` and `golangci-lint` both clean; `make
coverage` totals 88.2% against the 87% floor with every restructured
scheduler function covered; `make test` shows zero failures and zero
panics; the cache-key and extension-classification regression tests
(`TestCachedCheck_PerDatabaseVerdicts`,
`TestExecuteProbeForConnection_ExtensionPresentButEmpty/Absent`) all
pass; and the Codacy Lizard complexity findings on
`executeProbeForAllDatabases` are already resolved by the head commit's
own follow-up refactor.

**→ Merge order.** No textual or semantic conflict with #495 — confirmed
with `git merge-tree`, and the two PRs touch disjoint regions of
`executeProbeForConnection`. #493 and #464 do conflict for real in
`pg_stat_statements_probe.go` and its test (confirmed the same way);
your PR body already defers #439 there, so this isn't a surprise, but
the rebase will need to retire `featureCacheScope` in favour of the
`cachedCheck` signature change once #464 lands first.

PR #494, opened 2026-09-16
Report the queried time window from the metrics query endpoint

### LINKED ISSUE #430: Metrics query response omits the queried time window, so charts cannot show a range was honoured
## Description

`/api/v1/metrics/query` returns data points but never states which
time window it queried, so the client cannot distinguish "this metric
did not change over the selected period" from "no data exists for most
of that period". Charts therefore render identically across very
different selected ranges, which reads as the range selector being
broken.

This came out of investigating a user report that all stats looked the
same whether 1h or 7d was selected. Part of that report is the
hardcoded `time_range=24h` call sites tracked in #387, but this is a
separate and arguably more fundamental gap: even where the range is
plumbed through correctly, the UI cannot show that it was honoured.

## Cause

When a bucket has no data and there is no earlier value to carry
forward, the point is dropped rather than emitted:
`server/src/internal/metrics/query.go:958`, the `continue` in
`scanSeriesRows`. Buckets after the first sample are gap-filled last
observation carried forward; buckets before it produce nothing.

The response contains no window fields. `grep` for
`time_start`/`time_end` across `server/src/internal/metrics/` and
`server/src/internal/api/metrics_handlers.go` returns nothing, and the
handler parses the range at `metrics_handlers.go:134` without echoing
it back.

The client consequently derives its x-axis domain from the returned
points alone, so every chart begins at the first sample that exists
regardless of what was requested.

## Consequences

- On a newly deployed instance with a few hours of collection, 1h,
  6h, 24h, 7d and 30d all return the same handful of points and render
  pixel-identical charts on the same auto-scaled axis.

- The `30d` option in `TimeRangeSelector.tsx:26` can never show 30
  days for `pg_sys_*` or `pg_stat_*` probes, whose seeded retention is
  7 days (`collector/src/database/schema.go:512`). Because expiry
  drops whole Monday-aligned weekly partitions only once the
  partition's end precedes the cutoff (`collector/src/probes/partition.go:31`
  and `:150`), retained history slides between 7 and 14 days, so a 30d
  view sometimes differs from 7d and sometimes does not, with nothing
  on screen explaining why.

- LOCF-filled buckets are indistinguishable from real samples, so a
  chart of a 600s-interval probe over a 1h window presents ~144
  synthetic points and ~6 real ones with no visual difference.

## Proposed fix

Return the queried window, and ideally the data's own extent, from
`/api/v1/metrics/query`:

- `time_start` and `time_end` as requested, so the client can fix the
  x-axis domain to the selected range rather than to the data.

- The bucket width, which the client currently has to infer.

- Per series, the timestamp of the first and last real sample, or a
  per-point flag distinguishing observed values from carried-forward
  ones, so charts can render filled regions differently and avoid
  implying continuous observation.

The client should then anchor the axis to the requested window and
indicate periods with no data, so that an empty stretch reads as
absent history rather than as a flat metric.

Consider separately whether `30d` should be offered at all given
7-day retention, or whether the selector should be bounded by the
retention of the probes backing the visible charts.

## Related

- #387 hardcoded `time_range=24h` in the summary tiles and query views.
- #388 open PR adding custom start/end selection, which will inherit
  this gap: an arbitrary window is even harder to interpret without
  knowing what was actually queried.
- #400, #401, #402 cumulative counters and rate derivation, which are
  independent reasons charts look unchanging.

## Acceptance criteria

- The metrics query response states the window it queried and the
  bucket width.

- Charts anchor their x-axis to the selected range, so changing the
  range visibly changes the axis even when no additional data exists.

- Periods with no collected data are visually distinct from periods
  where a metric was flat.

- Tests cover the server response change and the client axis handling
  to the project's 90% line coverage floor.

dpage:
## Summary

The bucketed mode of `/api/v1/metrics/query` never said which window it
queried, so charts derived their x-axis from the returned points and every
range rendered identically on an instance with little history.

- The response is now an object: `probe_name`, `connection_ids`,
  `time_range`, `time_start`, `time_end`, `bucket_seconds`, `buckets`,
  `aggregation` and `series`. `bucket_seconds` comes from a shared
  `BucketWidth` helper that also backs the query builders, so the reported
  width is the width the SQL binned by. This is a breaking change to that
  response; the latest-rows mode (`limit`/`order_by`) is untouched.
- A carried-forward point is marked `filled`. Null values for buckets with
  no data already landed with the fill-policy work in #402.
- Dashboard charts build their categories from the returned window rather
  than from the data, so changing the range changes the axis; null buckets
  draw as gaps and each carried-forward run is shaded, drawn with a hollow
  marker and called out in the tooltip, so none of the three cues is colour
  alone.

Whether `30d` should be offered at all given seven-day probe retention is
deliberately left out: with the axis anchored, a 30d view now honestly shows
the missing history, and bounding the selector needs a retention endpoint the
client does not have.

## Test plan

- [x] `server`: `make test` and `make lint` clean; `BucketWidth` pinned to the
      SQL interval argument, and envelope tests cover preset and custom
      windows, a leading null bucket, and observed versus carried points.
- [x] `client`: 4136 tests pass, lint clean, no new typecheck errors;
      coverage above the ratchet on all four thresholds.
- [ ] Worth eyeballing on the dev host: switch a server dashboard between 1h,
      6h, 24h, 7d and 30d on an instance with a few hours of history and check
      the axis extent changes each time, that the shaded carried-forward bands
      read well in both themes, and that Cache Hit Ratio and Replication Lag,
      which take different code paths, anchor the same way.

Closes #430

## Summary by CodeRabbit

- **New Features**
  - Metrics queries now include time-window and bucket metadata alongside metric series.
  - Carried-forward values are identified separately from observed values.
  - Dashboard charts align to the reported time window, preserve unavailable periods as gaps, and visually distinguish carried-forward ranges with markers, shading, and tooltip labels.
  - Existing latest-rows behavior remains unchanged.
- **Documentation**
  - Updated API and dashboard documentation to describe response metadata, null values, bucket behavior, and carried-forward points.
- **Bug Fixes**
  - Improved chart alignment and handling of missing or out-of-range metric data.


AntTheLimey (approved):
No blocking findings, backend or client — this covered the widest
client surface in the current backlog (five dashboard types) and it
held up clean throughout.

The response envelope, the `BucketWidth` helper, and the `filled` flag
are all correctly wired: `buckets` and `bucket_seconds` reflect the
actual, interval-clamped values the SQL used (confirmed with
`TestQueryTimeSeries_Envelope`'s "clamped to the interval" case), and
`time_start`/`time_end` echo the resolved window exactly as requested,
by design — the PR description is explicit that a 30-day view over a
young instance should honestly show the missing history rather than
silently narrow to available data, and that decision reads correctly
end to end.

**→ Optional: raise `COVERAGE_MIN` in `server/Makefile`.** It is
unchanged at `59.9` while measured total coverage moved from **60.0%**
on `origin/main` to **60.1%** on this branch — a real, if small, lift
that #489 and #492 both ratcheted for comparable movement. Not a
blocker; the floor still passes with margin.

Everything else checks out: `make openapi` regenerates
`docs/admin-guide/api/openapi.json` byte-for-byte identical to what's
committed, the OpenAPI test suite passes (6 PASS, 1 environment skip),
`docs/admin-guide/api/reference.md` matches the schema including the
generate_series inclusive-bounds note, `gofmt -l` is clean,
`golangci-lint` reports 0 issues, `-race` is clean on both touched
packages, every touched function clears the 90% coverage floor
(`BucketWidth` and `handleMetricsQuery` both above 97%), and the
`.claude/golang-expert/metrics-queries.md` update is repo-specific
fact rather than the general-advice re-inflation #444 ruled against.
No open review threads need addressing (CodeRabbit hit its rate limit
and produced no findings; Codacy reports zero issues).

On the client, the window wiring is complete and consistent across
every dashboard type that consumes metrics, and the empty-window case
on a young instance is handled by design with a direct regression test
for exactly that scenario, not left to accident. Typecheck and lint
hold with zero regression, and every touched file clears the coverage
floor.

Merge order: one textual conflict against #489, both inserting a new
`Related Issues` bullet at the same spot in
`.claude/golang-expert/metrics-queries.md` — trivial to resolve by
keeping both bullets, whichever PR rebases second. No semantic overlap
between the two changes.

Issue #400, opened 2026-08-11
Dashboard charts plot raw cumulative counters instead of rates


dpage:
## Description

Roughly a third of the plotted series across the dashboards are raw
`pg_stat_*` counters sampled since `stats_reset`. They render as
monotonic ramps whose slope is dominated by how long the server has
been up rather than by current workload, so they cannot answer the
question a DBA is actually asking.

This was raised as user feedback: the charts show cumulative stats
that are not useful for diagnosing current activity.

## Affected charts

These plot counters raw and should show a per-second rate:

| Chart | File | Current fields |
|---|---|---|
| Transactions (Server) | `client/src/components/Dashboard/ServerDashboard/PostgresOverviewSection.tsx:365` | `xact_commit`, `xact_rollback` |
| Block I/O | `PostgresOverviewSection.tsx:394` | `blks_hit`, `blks_read` |
| Tuple Operations | `PostgresOverviewSection.tsx:423` | `tup_fetched/inserted/updated/deleted` |
| WAL Activity | `WalReplicationSection.tsx:318` | `wal_bytes`, `wal_records` |
| Network I/O | `SystemResourcesSection.tsx:529` | `tx_bytes`, `rx_bytes` |
| Transactions (Database) | `DatabaseDashboard/PerformanceSection.tsx:419` | `xact_commit`, `xact_rollback` |
| Calls Over Time | `ObjectDashboard/QueryDetail.tsx:710` | `calls` |
| Checkpoint buffers | `WalReplicationSection.tsx:377` | `buffers_written` |

The matching KPI sparklines carry the same defect and should change
alongside their parent charts: Commits and Temp Bytes
(`PostgresOverviewSection.tsx:287,314`), WAL Bytes, WAL Records and
Checkpoints (`WalReplicationSection.tsx:252,267,296`), and
Transactions (`PerformanceSection.tsx:382`).

## Rare-event metrics want a delta, not a rate

Per-second is the wrong presentation where the probe interval is 300
to 600 seconds and the event is rare; the value renders as 0.0011/s
and reads as zero. These should show a per-bucket delta drawn as bars:

- Checkpoint counts `num_timed` and `num_requested`
  (`WalReplicationSection.tsx:377`). The more useful derived signal is
  the requested fraction of total checkpoints; a high fraction means
  `max_wal_size` is too low.

- Temp bytes (`PostgresOverviewSection.tsx:314`), better expressed as
  megabytes spilled per window.

[...]

PR #496, opened 2026-09-16
Plot one real filesystem in the Disk Space chart and KPI tile

### LINKED ISSUE #428: Disk Space chart averages across all mounted filesystems
## Description

The Disk Space chart and its KPI tile average `used_space` and
`free_space` across every mounted filesystem, so the plotted values
correspond to no real filesystem and the chart cannot answer which
disk it refers to. This was raised as user feedback: on a host with a
dozen or more mounts, including `devtmpfs`, several `tmpfs` mounts and
separate volumes for data, WAL, logs and archive, the single averaged
line is not interpretable.

The stored data is correctly per-mount; only the chart blends it.

## Cause

The collector inserts one row per mounted filesystem returned by
`pg_sys_disk_info()`, with no filtering of pseudo filesystems:

- `collector/src/probes/pg_sys_disk_info_probe.go:40` selects
  `mount_point`, `file_system`, `total_space`, `used_space`,
  `free_space` and the inode columns.

- `collector/src/database/schema.go:1661` defines
  `metrics.pg_sys_disk_info` with primary key
  `(connection_id, collected_at, mount_point)`.

The client then requests those columns with `aggregation: 'avg'` and
no mount filter:

- `client/src/components/Dashboard/ServerDashboard/SystemResourcesSection.tsx:155`
  (chart) and `:113` (KPI tile).

The metrics query layer has no mount dimension at all, so every
filesystem collapses into one bucket:

- `server/src/internal/metrics/query.go:420` groups only by
  `date_bin($1::interval, collected_at, $3)`.

- `server/src/internal/metrics/query.go:339` emits
  `avg(used_space)` for `aggregation: 'avg'`.

- `MetricFilters` at `server/src/internal/metrics/query.go:50`
  supports only database, schema, table and index names.

The KPI tile compounds this by deriving a percentage from the
averages at `SystemResourcesSection.tsx:235`, which will not match any
real filesystem whenever mounts differ in size.

Note that `aggregation: 'last'` would not fix it either; the
`array_agg(... ORDER BY collected_at DESC)[1]` form at
`query.go:343` picks an arbitrary mount when several share a
timestamp.

## Paths that already handle mounts correctly

These are useful references for the fix, and are the current
workaround for anyone needing accurate figures:

- The Server Info dialog returns and renders one entry per mount:
  `server/src/internal/api/server_info_handlers.go:400` and
  `client/src/components/ServerInfoDialog/sections/SystemSection.tsx:106`.

- The estate query for Ellie's connection context is per-mount:
  `server/src/internal/database/estate_queries.go:792`.

- The alerter partitions by mount and takes the worst one, which is
  the right semantics for a threshold:
  `alerter/src/internal/database/metric_registry.go:941`.

## Proposed fix

Add `mount_point` as a filter and grouping dimension in the metrics
query path, then either let the user select a mount in the chart or
plot one series per real filesystem. Pseudo filesystems should be
excluded by default; `file_system_type` is already collected and is
the natural discriminator.

The KPI tile should report a specific filesystem, most usefully the
one closest to full, matching what the alerter already does.

## Related

Issue #400 lists the disk chart under "charts confirmed correct, for
reference"; that assessment is wrong and should be disregarded when
that issue is worked.

## Acceptance criteria

- The Disk Space chart identifies which filesystem or filesystems it
  is plotting, and no series is an average across unrelated mounts.

- Pseudo filesystems are excluded from the default view.

- The disk KPI tile reports a real filesystem's usage percentage.

- Tests cover the changed server query path and client components to
  the project's 90% line coverage floor.

dpage:
## Summary

The Disk Space chart and the Disk Usage KPI tile averaged `used_space`
and `free_space` across every mounted filesystem, so on a host with a
dozen mounts the plotted values corresponded to no real filesystem and
the chart could not say which disk it meant. The stored data was
always correctly per-mount; only the presentation blended it.

- `GET /api/v1/metrics/query` gains a `mount_point` filter, applied to
  both the time-series and the latest-row responses. The latest-row
  path builds its own `WHERE` rather than sharing the bucketed one, so
  it needed the clause separately; without it the parameter would have
  been dropped on exactly the path the dashboard uses to discover its
  mounts. The two handlers' identical filter literals moved into a
  shared `parseMetricFilters`.

- The chart and tile now describe one mount, defaulting to the
  filesystem closest to full (matching what the alerter reports), with
  a selector above the chart when there is more than one and the mount
  point in both the tile label and the chart title. Pseudo filesystems
  are excluded. Neither query runs until the mount is known, so the
  averaged query is unreachable rather than shown briefly on first
  paint, and a host with no real filesystem gets an empty state that
  says so rather than blaming `system_stats`.

- The usage percentage now comes from the filesystem's reported total
  rather than from used plus free, which disagree on any filesystem
  with reserved blocks.

- The alerter's `pg_sys_disk_info.used_percent` took the highest usage
  across all mounts, which is right for a threshold but pinned it at
  100% on any host with a snap installed, since a squashfs image is by
  construction fully used. It now ignores pseudo filesystems too.

Issue #400 lists the disk chart under "charts confirmed correct"; that
assessment was wrong and should be disregarded when #400 is worked.

## Test plan

- [x] `make -C server test`, `make -C alerter test`, `make -C client test`, run serially: all green, 4129 client tests passing.
- [x] Coverage gates: server 60.1% (min 59.9%), alerter 71.1% (min 70%), client lines 91.1%. Touched units are `metricQueryClauses` 100%, `buildLatestRowsQuery` 100%, `parseMetricFilters` 100%, `handleLatestRows` 100% (was 70.4%), `diskMounts.ts` 100%, `SystemResourcesSection.tsx` 95.0%.
- [x] `golangci-lint` 0 issues in both Go sub-projects; client lint 0 errors.
- [ ] Visual check in the browser, which needs the dev server running.

<details>
<summary>A note on the client coverage ratchet</summary>

The client thresholds move from 91.0/89.8/87.0/78.71 to
91.05/89.88/87.10/78.80, set a little under the measured totals rather
than exactly at them: two consecutive runs on a loaded host differed by
a hundredth of a point on three of the four figures, and a ratchet
pinned to the higher run fails the next one.
</details>

Closes #428


AntTheLimey (approved):
No blocking findings, backend or client.

**→ The `mount_point` dimension is correctly threaded through both query
paths.** `parseMetricFilters` in `metrics_handlers.go` is now the single
construction site for `metrics.MetricFilters`, so the parameter reaches
both `handleMetricsQuery` and `handleLatestRows`; `metricQueryClauses`,
`buildLatestRowsQuery` and the OpenAPI spec all pick it up in the same
change, verified with `cd server && make openapi` producing a
byte-identical `openapi.json` and
`TestBuildOpenAPISpec_MetricsQueryMountPointParam` passing.

**→ The per-mount identity is genuine, not a repeatable label.**
`metrics.pg_sys_disk_info` carries `PRIMARY KEY (connection_id,
collected_at, mount_point)` at the collector, so two different volumes
cannot share a `mount_point` and collapse into one summed row the way
#481's per-database bug did. `buildLatestRowsQuery`'s `DISTINCT ON`
picks `mount_point` up automatically as an entity key even when the
filter is absent, and the unfiltered-control test proves it returns one
row per real mount.

**→ On the client, the fix is correctly delegated to the server rather
than aggregated locally.** `SystemResourcesSection.tsx` picks one mount
(defaulted to the fullest real filesystem) and passes it as a single
`mount_point` param; both the KPI tile and the chart read the one
series that comes back. No client code sums or averages disk rows
itself.

**→ Optional: a mount-lookup fetch failure looks the same as an empty
host.** `useDiskMounts` returns an empty list either way, so the panel
shows "No real filesystem reported" for both a genuine network error
and a host with nothing to report. Matches an existing pattern
elsewhere in the same file, so it's consistent debt, not new.

**→ Everything measured is clean.** `cd server && make coverage`
against a live Postgres exits 0 at 60.1% (floor 59.9%), and `cd alerter
&& make coverage` exits 0 at 71.1% (floor 70%); both match the numbers
in the PR body exactly. The touched functions
(`parseMetricFilters`, `handleLatestRows`, `metricQueryClauses`,
`buildLatestRowsQuery`) are all at 100%, `handleMetricsQuery` at 96.9%.
`gofmt -l` and `golangci-lint run` are clean on every touched backend
package. No CodeRabbit review landed (rate-limited) and Codacy reports
zero issues, so there are no open review threads to address. The
knowledge-base updates in `metrics-queries.md` and
`quality-checklist.md` are repo-specific facts this change makes true,
not the general re-inflation #444 ruled against. The issue #400
correction in the PR description is accurate: #400 called the disk
chart "confirmed correct" on counter-vs-gauge grounds, but this PR shows
a real defect existed for an unrelated reason, so the correction is
worth keeping in #400 as a reference.

**→ Merge order: no collector files are touched**, so this PR sits
outside the four-way `migration_v12` collision affecting #468, #476,
#481 and #492.

PR #507, opened 2026-09-17
Move the alerter's default Gemini reasoning model to gemini-3.6-flash

### LINKED ISSUE #459: Alerter defaults its Gemini reasoning model to gemini-2.5-flash, which Google no longer offers to new keys
## Summary

The alerter's default Gemini reasoning model is `gemini-2.5-flash`, set in `alerter/src/internal/llm/reasoning.go` and `alerter/src/internal/config/config.go` and documented in `docs/getting-started/configuration/alerter.md` and `docs/developer-guide/alerter/anomaly-detection.md`. Google now refuses that model for new API keys:

```
Gemini API error (404): This model models/gemini-2.5-flash is no longer available to new users. Please update your code to use models/gemini-3.6-flash for the latest features and improvements.
```

So a fresh Gemini deployment that leaves `reasoning_model` unset fails on every reasoning call until the operator sets a model explicitly.

## Suggested change

Move the default to a current Gemini 3.x text model (Google's own message points at `gemini-3.6-flash`), update both code sites and both documentation pages, and add a changelog entry noting the behaviour change for existing users who relied on the implicit default. Worth checking the server's and `pgedge-go-llm-lib`'s own Gemini defaults for the same problem whilst there; the library deliberately left its default alone when fixing pgEdge/pgedge-go-llm-lib#35.

Split out from #425, where the 404 was reported alongside the thought-signature failure fixed in #456.

dpage:
Closes #459.

**Symptom.** A fresh Gemini deployment that leaves `llm.gemini.reasoning_model` unset fails on every reasoning call with `Gemini API error (404): This model models/gemini-2.5-flash is no longer available to new users`.

**Cause.** The alerter's implicit default was `gemini-2.5-flash`, which Google now refuses for new API keys.

**Change.** The default is now `gemini-3.6-flash` in `alerter/src/internal/llm/reasoning.go` and `alerter/src/internal/config/config.go`, with the two documentation pages, `examples/ai-dba-alerter.yaml` and `examples/walkthrough/guide.sh` updated to match, plus a changelog entry. The identifier was confirmed against Google's model list (https://ai.google.dev/gemini-api/docs/models): `gemini-3.6-flash` is a stable text-output model released 21 July 2026 with no shutdown date, and it is the model Google's own error message directs users to.

**Behaviour change.** An existing deployment that relied on the implicit default will start using `gemini-3.6-flash` after upgrading. Operators who need to stay on `gemini-2.5-flash` should set `reasoning_model` explicitly.

The server has no Gemini default of its own (its default provider is Anthropic and the model is always operator-supplied), and `pgedge-go-llm-lib` v0.3.1 has no default Gemini model either, so neither needs a matching change.

## Summary by CodeRabbit

- **Changed**
  - The default Gemini reasoning model is now `gemini-3.6-flash`.
  - Existing deployments that require `gemini-2.5-flash` must configure it explicitly.

- **Documentation**
  - Updated configuration examples, setup guidance, and walkthroughs to reflect the new default model.


AntTheLimey (approved):
This is a clean, self-consistent default-value swap and I'd approve it as is.

**→ Both external-dependency claims check out.** The server sets no Gemini `Model` default (only Ollama and Anthropic get one in `config.NewConfig`), and `pgedge-go-llm-lib` v0.3.1's `gemini.New()` takes `Model` straight from the caller with no fallback in `Options.WithDefaults()`. Neither needed a matching change.

**→ The replacement identifier is corroborated inside the repo, if not independently by me.** Issue #459 quotes the real Google 404 text pointing at `models/gemini-3.6-flash`, which lines up with the PR body's claim. I have no way to hit Google's live API from this review to confirm `gemini-3.6-flash` itself stays healthy going forward, but that is inherent to this kind of change, not something specific to flag here.

**→ Tests, coverage and lint all pass.** `TestNewConfig` asserts the new default directly, `gofmt`/`golangci-lint` are clean, and `make coverage` lands at 77.0% against the 76.0% floor with both touched functions at 100% line coverage. One optional gap: no test drives Gemini's own empty-model fallback through `newLibReasoning` the way the `openai` and `anthropic` cases do, though the production path that matters (`config.NewConfig`'s default) is covered. Docs, the example config and the walkthrough script all read `gemini-3.6-flash` consistently, and a repo-wide grep for the old value outside the touched files turns up only inert test literals in the server package.

PR #511, opened 2026-09-17
Fix empty sparkline canvas in Database Summaries

### LINKED ISSUE #458: Database Summaries cache hit sparklines render an empty canvas
The cache hit ratio sparklines in the Server Dashboard's Database
Summaries section never draw anything. The canvas is created at the
right size and stays completely empty, so every row shows a numeric
ratio above a blank strip where the trend should be.

This is not a data problem: the same rows display a live `current`
value, and the guard that renders the sparkline only passes when
`time_series.length > 0`.

## Evidence

Measured during a visual review pass on 2026-09-12, against the
production client build in the compose stack.

- Both canvases are present and correctly sized at 275x30.
- Reading the canvas backing store gives **0 of 8250 pixels
  non-transparent**, under idle and under pgbench load (99.9% hit
  ratio, 19.7 txn/s).
- Still empty after forcing a resize, so it is not a first-paint
  sizing race.
- Larger charts on the same page, fed from the same API response,
  render normally.

## Reproduce

1. Register a monitored server and let the collector gather at least
   two cycles of `pg_stat_database`.
2. Open the Server Dashboard and expand **Database Summaries**.
3. Each row shows a cache hit percentage with an empty 30px strip
   beneath it.

## Where to look

`client/src/components/Dashboard/ServerDashboard/DatabaseSummariesSection.tsx`
is the only place that renders `<Sparkline>` directly. Every other
sparkline on these dashboards goes through `KpiTile`, and those all
render.

The two call sites differ in exactly two ways, which is where I would
start:

- This one passes `height={30}`; `KpiTile` uses the default 40.
- This one wraps the chart in `CHART_BOX_SX`, a hard `height: 30`.
  `Chart` renders into a `Paper` carrying `CHART_PAPER_SX` with
  `p: 2` and `overflow: 'hidden'`, so the chart needs roughly 62px of
  box to place a 30px canvas. `KpiTile` gives it a flexible
  `{ mt: 1, flex: 1, minHeight: 0 }` container instead.

I have not confirmed that geometry mismatch is the cause, only that it
is the difference between the broken call site and the working ones.

## Notes

Pre-dates #450; it is not a regression from the recent metrics work.
Screenshots from the review pass are
`database-summaries-sparklines-active-load-empty.png` and
`database-summaries-sparklines-prepgbench-idle-values.png`.

dpage:
## Cause

Two independent faults, both of which have to be fixed before a
Database Summaries row draws anything.

**The plot area collapses.** `Sparkline` overrides the chart grid insets,
but `Chart` deep-merges its `echartsOptions` over `buildGrid()`'s options
key by key, so `containLabel: true` survives. That layout pass reserves
room for each axis's estimated label rect whether or not the axis is
shown, which at sparkline heights leaves nothing to draw into.

**A series with no adjacent pair of values has no geometry.** ECharts
draws a line series as segments between neighbouring values, so a single
point, or points separated by null gaps, paint nothing at all once
markers are off: the area path degenerates to a zero-width `M x y L x y
Z` and the line path to a bare move. Issue #458's own repro leaves
exactly one point in `time_series`, and so does a newly registered
server, which is the most common way to hit this.

Measured in Chrome against **echarts 6.1.0** (the version
`package-lock.json` pins) on a 275x30 canvas, through the real option
pipeline, counting non-transparent canvas pixels:

| series | `containLabel` | markers | grid rect | pixels drawn |
|--------|----------------|---------|-----------|--------------|
| 12 points | true | off | 235.8 x **-6.1** | 708 / 8250, in a 3px band |
| 12 points | false | off | 271 x 26 | 7220 / 8250 |
| 1 point | false | off | 271 x 26 | **0 / 8250** |
| 1 point | false | on | 271 x 26 | 69 / 8250 |
| `[null, v, null, v]` | false | off | 271 x 26 | **0 / 8250** |
| `[null, v, null, v]` | false | on | 271 x 26 | 104 / 8250 |

The same collapse hits `KpiTile`'s default 40px canvas, which came out
as a 5px band rather than the full 37px, so it was squashed rather than
visibly empty. The `CHART_BOX_SX` geometry mismatch noted on the issue
is real but is not the cause, so it is left alone.

An earlier revision of this PR measured the first row on echarts 6.0.0
and reported a 4px plot rect; on 6.1.0 the rect height is negative and
the drawn band is 3px. The table above replaces those figures.

## Change

`Sparkline` sets `grid.containLabel: false`, and passes `showMarkers`
to `Chart` when the series holds no adjacent pair of values, so a lone
or fully gapped series gets a circle on each observation and turns the
markers off again as soon as the line can be joined up. A wholly empty
series, or one where every bucket is a null gap, still renders no chart
at all: there is nothing to show, and an empty canvas would be worse
than no canvas.

The `axisLabel: { show: false }` overrides the first revision added are
dropped. They were redundant: `containLabel: false` skips the label
reservation outright, and the grid rect is identical with and without
them. The code comment and the `.claude/react-expert` entry that
claimed all three flags were necessary are corrected.

## Tests

The existing regression tests mock `Chart` wholesale, so they assert on
prop shape and could not have caught either fault. `Sparkline.render.
test.tsx` mocks only the `echarts-for-react` wrapper and runs the fully
merged option object through ECharts' own SSR renderer, then measures
the subpaths actually emitted; all four of its geometry assertions fail
against the pre-fix component. The prop-shape tests stay as the cheap
guard on the options themselves.

## Test plan

- [x] `make lint`: 0 errors, 37 pre-existing warnings, none in the touched files.
- [x] `npx tsc --noEmit`: clean.
- [x] `make coverage`: 4346 passed; `Sparkline.tsx` 100% statements and branches; all four client ratchets clear (lines 91.68, statements 90.54, functions 88.14, branches 79.96).

Closes #458

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dashboard sparklines at small heights by preserving more plotting space.
  - Ensured isolated or gap-separated data points remain visible.
  - Maintained hidden axes and prevented rendering when all data points are unavailable.

- **Tests**
  - Added coverage for single-point, gapped, multi-point, empty, and unavailable data series.
  - Added validation of the rendered chart output and layout behavior.


AntTheLimey (changes requested):
**This fix does not close #458 as reproduced, and the root-cause measurement table does not hold on the echarts version this project ships.**

**→ Single/short-history data is still blank after the fix.** Issue #458's own repro ("at least two cycles" of `pg_stat_database`) leaves exactly one `time_series` point. Feeding the PR-head build (`cfc058bf`) that exact shape through the real API path gives a Cache Hit Ratio canvas with 0 of 6630 non-transparent pixels — the reported symptom, unchanged. That is also the state a newly registered server is in, which is likely the most common real-world trigger.

**→ The claimed containLabel collapse does not reproduce on echarts@6.1.0, which is what `package-lock.json` pins** (the PR's table was measured against 6.0.0). Rebuilding the unmodified pre-fix `Sparkline.tsx` and feeding it a realistic 12-point series (matching the real API's shape) renders a sparkline with a non-transparent bounding box 27px tall of the 30px canvas — not the 4px collapse described for `containLabel: true`.

**→ Recommend re-diagnosing before merging.** The evidence points at a different, unaddressed cause: a line/area series with one point has nothing to draw a segment between, independent of grid or axis-label settings. The grid change itself looks safe to keep, but closing #458 on it leaves the single-point case (the most likely real trigger) exactly as broken as before.

Two smaller points worth a look, not blockers: `grid.containLabel: false` already skips ECharts' label-space reservation per its own docs, which suggests the two `axisLabel: { show: false }` overrides may be redundant for this fix rather than each independently necessary as the comment and `quality-checklist.md` entry assert; and the regression tests mock `Chart` entirely, so they assert on prop shape only and could not have caught either finding above — worth a rendered or screenshot-level check for this bug class.

Everything else checks out: ESLint clean on both touched files, `Sparkline.tsx` at 100% statements/branches/functions/lines matching the stated claim, the full client suite passing with coverage above every pinned threshold, and the `quality-checklist.md` change is purely additive. No blocking review threads exist (CodeRabbit hit its rate limit before reviewing, Codacy is excluded from this review per house rules). PR #510 merges cleanly with this branch per the coordinator's check.

AntTheLimey (approved):
Both blockers from the last review are fixed, verified by rebuilding the stack and reproducing issue #458's exact single-point shape against this head, and independently re-confirmed with a deterministic test that fails on the pre-fix component and passes on this one.

The single-point case now renders. Mocking the API to return exactly one cache_hit_ratio.time_series point and pixel-scanning the Cache Hit Ratio canvas gives a visible marker where the previous head gave zero non-transparent pixels on the identical repro. A gap-separated series (null, 95, null, 98) also now renders two markers. A realistic 12-point series still fills the canvas.

I confirmed this a second way, independent of any screenshot: the PR's new Sparkline.render.test.tsx renders <Sparkline data={series([97])} height={30} /> — the exact real call site's configuration — through ECharts' own SSR renderer and checks what actually got drawn. On this head, 24/24 tests pass. Swapping in the pre-fix Sparkline.tsx with the same test file, the single-point and gap-separated tests fail immediately while the other four still pass. That's a clean, repeatable proof the fix works on the real component and the real configuration.

The root-cause table is now measured against the version this project ships. echarts@6.1.0 is confirmed installed and pinned in package-lock.json. The PR body states plainly that the earlier 6.0.0 table is superseded, with new figures for 6.1.0.

The axisLabel redundancy claim from the last review is corrected, not just asserted. Running the production option object through echarts@6.1.0's SSR renderer with and without the two axisLabel: { show: false } overrides, holding containLabel: false constant, gives an identical grid rect both times. The PR removes both overrides and corrects the code comment and the quality-checklist.md entry to match.

The test gap is closed. Sparkline.render.test.tsx renders the fully merged option object through ECharts' own SSR renderer instead of mocking Chart, and its assertions fail against the pre-fix component for exactly the two cases that mattered.

Checks: ESLint clean on the touched files (37 warnings, 0 errors across the whole client, none in this diff); tsc --noEmit clean; Sparkline.tsx at 100% statements/functions/branches; full suite at 4346 passed with all four coverage ratchets clear; branch is 0 behind origin/main and merges cleanly; CodeRabbit's full review and Codacy both report no issues.

Approving.

PR #516, opened 2026-09-21
Slim CLAUDE.md and move its reasoning next to the code

dpage:
`CLAUDE.md` is sent on every turn of every session, so its size is multiplied
by the number of turns and then again by the number of sessions. At 11KB it
was costing roughly 2.9k tokens per turn.

Nothing here is deleted. Each rule stays in `CLAUDE.md`; the prose explaining
or justifying it moves to the file a reader would have open at the moment
they needed it:

- the manual tooling fallback, into a comment at the end of
  `.claude/hooks/check-tooling.sh`
- the knowledge-base staleness rationale, into the three writing agents' own
  briefs, folded into the sentence that already stated the rule
- the `playwright-cli` installation procedure and the reason for validating
  UI changes in a browser, into that skill's own file
- the copyright notice, replaced by a pointer to the contributing guide,
  which already carries it

**Worth a second look:** that last move turned up a real discrepancy. The
contributing guide showed a 55-dash rule whilst all 196 Go files in the tree
use the 73-dash form, so anyone following the guide was producing a header
that did not match. The guide now matches the code, which is the one change
here that alters something beyond Claude's own instructions.

`CLAUDE.md` goes from 11,021 to 9,517 bytes. My private `CLAUDE.local.md` is
git-ignored and shrank alongside it, from 11,110 to 7,175, but none of that
appears here.

## Summary by CodeRabbit

- **Documentation**
  - Clarified contributor workflows, pull-request procedures, worktree practices, and knowledge-base synchronization requirements.
  - Added guidance for visually validating browser-based interface changes.
  - Documented browser testing setup, including installation requirements and a basic smoke test.
  - Updated copyright-header guidance and clarified handling of local plans, specifications, and notes.

- **Chores**
  - Improved tooling checks and fallback guidance for documentation and browser-related tasks.
  - Excluded local notes from version control.


dpage:
Merged with override. This is CLAUDE tuning only.

PR #527, opened 2026-09-21
Scope the CREATE TABLE convention to notification_channels

### LINKED ISSUE #504: Collector migration 12 skips the base table, against the documented convention
Noticed whilst landing #476.

`.claude/golang-expert/notification-channels.md` states that a column added by a
migration must also be added to the `CREATE TABLE` in `collector/src/database/schema.go`,
so that a fresh install gets the final shape from migration 1 rather than depending on a
later migration's `ALTER`/`DROP CONSTRAINT IF EXISTS` path, which is the more fragile of
the two.

Migration 12 from #384 (`query_id` on `metrics.pg_stat_activity`) does not do this: the
partitioned parent's `CREATE TABLE` has no `query_id` column, so fresh installs get it
only via the migration. #476 restored the equivalent base-table change for its own
migration 14, which is what surfaced the difference.

So the convention is currently applied inconsistently in the tree, and one of two things
should happen: either add the column to the base table for migration 12, or drop the
convention from the knowledge base and say plainly that migrations alone define the
final shape. A knowledge-base entry that the code contradicts is the outcome worth
avoiding either way.

dpage:
## Summary

#504 read the `CREATE TABLE` rule in `.claude/golang-expert/notification-channels.md` as project-wide and concluded that migration 12 (`query_id`) violates it. Checking the other migrations, the picture is the other way round:

- The rule is written about `CREATE TABLE notification_channels`, and every `notification_channels` migration follows it, including migration 14 from #476.
- The partitioned `metrics.*` tables have never followed it. Migrations 7 (`table_size`, `index_size`), 10 (`stats_reset`), 12 (`query_id`) and 13 (`available_memory`) all add their column by `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` on the partitioned parent alone, with nothing in the base `CREATE TABLE`.

So migration 12 is not an outlier, it is four of four, and there is no schema inconsistency to fix. What was wrong is the knowledge base implying one rule where the tree has two, each right for its own kind of table. This change makes both explicit: the `notification_channels` entry now says it governs that plain configuration table, and `partitioning.md` gains a section recording that for `metrics.*` the numbered migration is the only place a later column appears, which works because PostgreSQL propagates `ADD COLUMN` on a partitioned parent to existing and future partitions.

Documentation only: no schema, code or test changes.

## Test plan

- [x] Verified against `collector/src/database/schema.go` that none of `table_size`, `index_size`, `stats_reset` (on `pg_stat_statements`), `query_id` or `available_memory` appears in its base `CREATE TABLE`, and that the Telegram columns from migration 14 do appear in the `notification_channels` one.
- [ ] No test run applicable; knowledge-base files only.

Closes #504

## Summary by CodeRabbit

* **Documentation**
  * Clarified migration guidance for notification channel and partitioned metrics tables.
  * Added instructions for adding columns to partitioned tables after the initial migration.
  * Documented how PostgreSQL propagates schema changes to existing and future partitions.
  * Clarified that non-partitioned configuration tables require separate handling.


AntTheLimey (approved):
This is a correct, well-scoped fix and I'd merge it as-is.

**→ The migration claims check out exactly.** I read
`collector/src/database/schema.go` directly: migrations 7, 10, 12 and
13 each add their column only via `ALTER TABLE ... ADD COLUMN IF NOT
EXISTS` on the partitioned parent, none of those five columns appears
in migration 1's base `CREATE TABLE` blocks, and migration 13's own
comment already says the same thing about migrations 7 and 10.

**→ The scoping is accurate, not just plausible.** Migration 1's
`CREATE TABLE notification_channels` block already carries the
Telegram columns that migration 14 also adds by numbered migration, so
the original rule still holds for `notification_channels` — this PR
narrows an overclaimed rule rather than papering over a real
inconsistency, which matches what your own test plan found.

**→ Neither new passage overreaches.** The `partitioning.md` addition
is scoped to `metrics.*`, and its cross-reference to
`notification-channels.md` names `notification_channels` specifically
rather than generalizing to "configuration tables." I found no prior
history of similar content being trimmed from either file, so this
reads as new, repo-specific documentation rather than a re-inflation
of pruned general advice.

Both files stay within the 79-character wrap, the diff is
documentation-only with no code, test, OpenAPI or changelog surface to
update, and no other open PR touches either file, so there's no merge
ordering to worry about.

AntTheLimey (approved):
The new commit holds up and I'd keep the Approve.

**→ Both new facts check out.** Migration 2 adds `clusters.dismissed` and migration 5 adds `metric_baselines.earliest_sample_at`, both by `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` alone, and neither column appears in migration 1's base `CREATE TABLE` blocks for those tables.

**→ The list is now exhaustive, not just bigger.** I grepped every `ADD COLUMN IF NOT EXISTS` in a numbered migration in `schema.go`: the full set is `clusters` (2), `metric_baselines` (5), the four `metrics.*` columns (7, 10, 12, 13) already documented, and `notification_channels` (14, the stated exception). Nothing is missing and nothing extra crept in. `chat_memories.embedding` uses the same pattern but runs inside migration 1's own SAVEPOINT, not a later migration, so it correctly stays out of scope.

**→ The two files still agree with each other.** `partitioning.md`'s cross-reference continues to name only `notification_channels` as the exception, so neither file drifted out of sync with the other's claim.

The commit stays within the 79-character wrap, touches no code, test, OpenAPI, or changelog surface, and CodeRabbit's re-review of this specific commit found nothing actionable. PR #536 also edits `notification-channels.md` but not the paragraph this commit changes; `git merge-tree` shows a clean merge, so no ordering conflict.

